PR #5, opened 2022-07-09
Handle another error code when iouring initialization fails

iko1:


romange:
Hey @iko1 , thanks for fixing the issue. LGTM - I do not have any problem accepting this improvement but wanted to suggest something else. so choose whatever you prefer. I originally thought to try initializing a temporary iouring in dfly_main.cc in dragonfly repo right after I check for kernel version, this way you can print nicely in the main thread if your system requirements are not met. and close it in case of success of course.

util/uring/proactor.cc
```diff
@@ -428,6 +428,11 @@ void Proactor::Init(size_t ring_size, int wq_fd) {
              "memory is too limited. If you run me via docker, try adding '--ulimit memlock=-1' to"
              "docker run command";
       exit(1);
+    } else if (init_res == ENOSYS) {
+      LOG(ERROR)
+          << "io_uring system call interface is not support by current kernel. "
+             "Try to install kernel with io_uring support.";
+      exit(0);
```
romange, line 435: I would exit with an error. See above.

iko1:
> Hey @iko1 , thanks for fixing the issue. LGTM - I do not have any problem accepting this improvement but wanted to suggest something else. so choose whatever you prefer. I originally thought to try initializing a temporary iouring in dfly_main.cc in dragonfly repo right after I check for kernel version, this way you can print nicely in the main thread if your system requirements are not met. and close it in case of success of course.

Where should I place the include for the file `liburing.h` in dfly_main.cc?

romange:
Yes, of course.  I usually put c includes before c++ includes, so in this case it should be the first one.

iko1:
> Yes, of course. I usually put c includes before c++ includes, so in this case it should be the first one.

This code should be sufficient to check whether iouring is not supported:
`io_uring_queue_init_params(0, nullptr, nullptr);`
and then `errno `is equal to `ENOSYS`

I ran it on my Ubuntu 20.04 with unsupported iouring kernel version `5.10.43.3`. 
WDYT?

PR #14, opened 2022-09-24
Add support for AWS_SESSION_TOKEN

dranikpg:


util/cloud/aws.h
```diff
@@ -43,6 +44,7 @@ class AWS {
   void SetScopeAndSignKey();
 
   std::string region_, service_, secret_, access_key_;
+  std::optional<std::string> session_token_;
```
romange, line 47: you could get away with using an empty value instead of using `optional`. Why do you prefer optional? readability?

dranikpg: Yes, empty value will work for sure. Optional just clearly shows this is an optional value, feels somewhat cleaner. But I can change it back

romange (approved):

PR #17, opened 2022-09-29
feat(util): Support pagination in AWS ListObjects

### LINKED ISSUE #12: S3 object list api does not support paging
currently function `ListObjects` in s3.h check fails, if it encounters a truncated list of objects in the XML s3 response.
 
also the S3Bucket API does not support it. We must fix it before completing #10

dranikpg:
Signed-off-by: Vladislav Oleshko <vlad@dragonflydb.io>

Fixes #12

### DIFF
--- a/util/cloud/s3.cc
+++ b/util/cloud/s3.cc
@@ -15,6 +15,7 @@
 #include "base/logging.h"
 #include "util/cloud/aws.h"
 #include "util/proactor_base.h"
+#include "util/http/encoding.h"
 
 namespace util {
 namespace cloud {
@@ -26,7 +27,7 @@ using nonstd::make_unexpected;
 const char kRootDomain[] = "s3.amazonaws.com";
 
 // Max number of keys in AWS response.
-const int kAwsMaxKeys = 1000;
+const unsigned kAwsMaxKeys = 1000;
 
 inline std::string_view std_sv(const ::boost::beast::string_view s) {
   return std::string_view{s.data(), s.size()};
@@ -116,10 +117,10 @@ ListObjectsResult ParseListObj(string_view xml_obj, S3Bucket::ListObjectCb cb) {
   for (xmlNodePtr child = root->children; child; child = child->next) {
     if (child->type == XML_ELEMENT_NODE) {
       xmlNodePtr grand = child->children;
-      if (!strcmp(as_char(child->name), "IsTruncated")) {
-        truncated = strcmp(as_char(grand->content), "true") == 0;
-      } else if (!strcmp(as_char(child->name), "Marker")) {
-      } else if (!strcmp(as_char(child->name), "Contents")) {
+      if (as_char(child->name) == "IsTruncated"sv) {
+        truncated = as_char(grand->content) == "true"sv;
+      } else if (as_char(child->name) == "Marker"sv) {
+      } else if (as_char(child->name) == "Contents"sv) {
         auto sz_name = ParseXmlObjContents(child);
         cb(sz_name.first, sz_name.second);
         last_key = sz_name.second;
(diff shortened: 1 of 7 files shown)


util/cloud/s3.cc
```diff
@@ -25,6 +25,9 @@ using nonstd::make_unexpected;
 
 const char kRootDomain[] = "s3.amazonaws.com";
 
+// Max number of keys in AWS response.
+const int kAwsMaxKeys = 1000;
+
```
dranikpg, line 30: This could be moved to the header to be part of the signature

util/cloud/s3.cc
```diff
@@ -95,33 +98,36 @@ std::pair<size_t, string_view> ParseXmlObjContents(xmlNodePtr node) {
...
-
   for (xmlNodePtr child = root->children; child; child = child->next) {
     if (child->type == XML_ELEMENT_NODE) {
       xmlNodePtr grand = child->children;
       if (!strcmp(as_char(child->name), "IsTruncated")) {
-        CHECK(grand && grand->type == XML_TEXT_NODE);
-        CHECK_STREQ("false", as_char(grand->content)) << "TBD: need to support cursor based functionality";
+        truncated = strcmp(as_char(grand->content), "true") == 0;
       } else if (!strcmp(as_char(child->name), "Marker")) {
       } else if (!strcmp(as_char(child->name), "Contents")) {
```
dranikpg, line 122: `== "const"sv` looks cleaner, `== 0` is easier to understand. But I'll change it back to `truncated = !strcmp`

util/cloud/s3.cc
```diff
@@ -165,8 +171,20 @@ std::error_code S3Bucket::Connect(uint32_t ms) {
...
-  h2::request<h2::empty_body> req{h2::verb::get, "/", 11};
+ListObjectsResult S3Bucket::ListObjects(string_view bucket_path, ListObjectCb cb,
+                                        std::string_view marker, int max_keys) {
+  std::string path = absl::StrCat("/", bucket_path, "?");
+
+  if (marker != "")
+    path += absl::StrCat("marker=", marker, "&");
+
+  if (max_keys != kAwsMaxKeys)
+    path += absl::StrCat("max-keys=", max_keys, "&");
```
dranikpg, line 182: Too much url building code in one place

util/cloud/s3.cc
```diff
@@ -165,8 +171,20 @@ std::error_code S3Bucket::Connect(uint32_t ms) {
...
+
+  if (path.back() == '?' || path.back() == '&')
+    path.pop_back();
```
dranikpg, line 185: The If clause is always true. A better alternative would be:
```c++
CHECK(path.back() == '?' || path.back() == '&');
path.pop_back();
```

util/cloud/s3.cc
```diff
@@ -220,6 +238,20 @@ error_code S3Bucket::ListObjects(string_view path, ListObjectCb cb) {
   return xml::ParseListObj(xml, std::move(cb));
 }
 
+error_code S3Bucket::ListAllObjects(string_view bucket_path, ListObjectCb cb) {
```
dranikpg, line 241: 1. Decide on naming (List vs ListAll, Get vs List)
2. Maybe pass batch size as optional param

util/cloud/s3.cc
romange: i agree

util/cloud/s3.cc
romange: are those comments to you? to me? :)

util/cloud/s3.h
```diff
@@ -13,6 +13,9 @@ namespace util {
 namespace cloud {
 
 using ListBucketsResult = io::Result<std::vector<std::string>>;
+// Inner result value is 'marker' to start next page from.
```
romange, line 16: empty line before comments.

romange (commented):
looks good

util/cloud/s3.cc
```diff
@@ -165,8 +171,20 @@ std::error_code S3Bucket::Connect(uint32_t ms) {
   return http_client_->Connect(host, "80");
 }
 
-error_code S3Bucket::ListObjects(string_view path, ListObjectCb cb) {
-  h2::request<h2::empty_body> req{h2::verb::get, "/", 11};
+ListObjectsResult S3Bucket::ListObjects(string_view bucket_path, ListObjectCb cb,
+                                        std::string_view marker, int max_keys) {
```
romange, line 175: int -> unsigned or uint32_t

util/cloud/s3.cc
dranikpg: To you 😅 Imagine a question mark after each of them

dranikpg:
For some reason the signature is wrong when a parameter contains a slash (marker can) 🤔. I'll look into it.

By the way, AWS suggests using its newer ListObjectsV2

util/cloud/s3.cc
romange: i do not mind, whatever you prefer to do here.

util/cloud/s3.h
```diff
@@ -25,7 +28,10 @@ class S3Bucket {
   //! Called with (size, key_name) pairs.
   using ListObjectCb = std::function<void(size_t, std::string_view)>;
 
-  std::error_code ListObjects(std::string_view path, ListObjectCb cb);
+  ListObjectsResult ListObjects(std::string_view path, ListObjectCb cb,
+                                std::string_view marker = "", int max_keys = 1000);
+
+  std::error_code ListAllObjects(std::string_view path, ListObjectCb cb);
```
romange, line 34: add a comment saying that this a convenience function that handles pagination internally and iterates over all matching items.

util/cloud/s3.cc
romange: 1. Naming is fine (I think List is better here).
2. No need, sane defaults should work until they won't and then we will decide what to do.

romange (approved):

Issue #1, opened 2022-01-15
statx is not implemented by musl-dev


romange:
we should check for statx support and avoid using it if it does not exist

romange:
it's already supported in alpine

PR #21, opened 2022-10-25
fix: __sigval_t causes compile error.

Super-long:
Signed-off-by: Super-long <0x4f4f4f4f@gmail.com>

I encountered the following problem when compiling：
截屏2022-10-25 20 29 35

__sigval_t has limitations, to take two different examples：
Linux VM-9-54-centos 5.4.119-1-tlinux4-0008 #1 SMP Fri Nov 26 11:17:45 CST 2021 x86_64 x86_64 x86_64 GNU/Linux
截屏2022-10-25 20 15 01

Linux VM-16-16-ubuntu 5.4.0-126-generic #142-Ubuntu SMP Fri Aug 26 12:12:57 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
截屏2022-10-25 20 27 28

This change was made for better compatibility.

### DIFF
--- a/util/dns_resolve.cc
+++ b/util/dns_resolve.cc
@@ -25,7 +25,17 @@ struct NotifyStruct {
   fibers_ext::EventCount evc;
 };
 
-static void DnsResolveNotify(__sigval_t val) {
+// To ensure compatibility between different versions of Glibc,
+// we use sigval_t instead of __sigval_t. However, some older 
+// versions may still require __sigval_t, such as when __USE_POSIX199309
+// is defined. The following text is derived from the comments in Glibc 2.31:
+// To avoid sigval_t (not a standard type name) having C++ name
+// mangling depending on whether the selected standard includes union
+// sigval, it should not be defined at all when using a standard for
+// which the sigval name is not reserved; in that case, headers should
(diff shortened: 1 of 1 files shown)


romange:
Looks good, can you please add a comment saying that `__sigval_t` is not compatible with some linux distributions? Otherwise I will break it again at some point.

Super-long:
@romange Of course, this may take a bit of time, but I will add this information soon.

Super-long:
This continues to look like a Glibc version issue, mentioned in this document: **sigval_t, although not a standard name, seems to be widely used outside of glibc.**
https://patchwork.ozlabs.org/project/glibc/patch/alpine.DEB.2.20.1708111812040.2884@digraph.polyomino.org.uk/

There is reason to believe that the new Glibc version will be compatible with this, and I also found the following code in glibc2.31:
截屏2022-10-30 18 53 59

So we should change __sigval_t to sigval_t.

romange:
I agree. Can you just add a comment before `DnsResolveNotify` definition that says we must use `sigval_t` in order to support differrent glibc distributions besides ubuntu.

Super-long:
Of course, I added some comments to explain this change.

romange (approved):

PR #23, opened 2022-10-28
Improve OnIdleTasks management

romange:
Signed-off-by: Roman Gershman <romange@gmail.com>

### DIFF
--- a/util/proactor_base.cc
+++ b/util/proactor_base.cc
@@ -82,53 +82,63 @@ void ProactorBase::Stop() {
   VLOG(1) << "Proactor::StopFinish";
 }
 
-uint32_t ProactorBase::AddIdleTask(IdleTask f) {
+uint32_t ProactorBase::AddOnIdleTask(OnIdleTask f) {
   DCHECK(InMyThread());
 
-  auto id = next_task_id_++;
-  auto res = on_idle_map_.emplace(id, std::move(f));
-  CHECK(res.second);
-  idle_it_ = on_idle_map_.begin();  // reset position to the first item.
+  uint32_t res = on_idle_arr_.size();
+  on_idle_arr_.push_back(OnIdleWrapper{.task = std::move(f), .next_ts = 0});
 
-  return id;
+  return res;
 }
 
-void ProactorBase::RunOnIdleTasks() {
-  if (on_idle_map_.empty())
-    return;
-
-  if (idle_it_ == on_idle_map_.end()) {
-    idle_it_ = on_idle_map_.begin();
-  }
+bool ProactorBase::RunOnIdleTasks() {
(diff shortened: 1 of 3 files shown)


util/proactor_base.cc
```diff
@@ -66,6 +66,8 @@ ProactorBase::ProactorBase() : task_queue_(512) {
 }
 
 ProactorBase::~ProactorBase() {
+  on_idle_arr_.clear();
+
```
dranikpg, line 70: Doesn't this happen on destruction either way 😅

util/proactor_base.cc
```diff
@@ -82,53 +84,63 @@ void ProactorBase::Stop() {
...
 
-void ProactorBase::CancelIdleTask(uint32_t id) {
-  auto it = on_idle_map_.find(id);
-  if (it != on_idle_map_.end()) {
-    on_idle_map_.erase(it);
-    idle_it_ = on_idle_map_.begin();
-  }
+bool ProactorBase::RemoveOnIdleTask(uint32_t id) {
+  if (id < on_idle_arr_.size() || !on_idle_arr_[id].task)
+    return false;
```
dranikpg, line 139: The `if  id < on_idle_arr_.size() then return` seems odd. I guess it should be the opposite

util/proactor_base.cc
```diff
@@ -82,53 +84,63 @@ void ProactorBase::Stop() {
...
-    if (!res) {
-      on_idle_map_.erase(idle_it_);
-      idle_it_ = on_idle_map_.begin();
-      break;
+    if (on_idle.task && on_idle.next_ts <= curr_ts) {
+      tl_info_.monotonic_time = curr_ts;
+
+      uint32_t level = on_idle.task();  // run the task
+
+      curr_ts = GetClockNanos();
```
dranikpg, line 116: If `GetClockNanos` is non monotonic, then you might end up jumping back in time and even running the same task twice (if it has a high level and a small delta). Or is that highly unlikely? So a `curr_ts = max(curr_ts, GetClockNanos())` would solve it but probably bring other problems. Maybe its even not important at all

util/proactor_base.cc
romange: yeah, I guess you are right.

util/proactor_base.cc
romange: yeah, this is not what worries me. as long as we won't have weird integer overflow i do not care.
the corner case is when someone changes the system clock. we will need to test it some day anyway.

util/proactor_base.cc
romange: good catch

romange:
@dranikpg ptal

dranikpg (approved):

PR #27, opened 2022-11-23
mimalloc patching for memory defrag

boazsade:
This PR is for enabling memory defragmentation using mimalloc
There are two issue that this should solve:
1. We are building mimalloc as release version, but DF can then build as debug, this create different structs sizes between mimallc and DF (because of different flags that the header files and source files "see").
2. Adding a function that checks whether a given page is underutilised.
3. the third party cmake will apply the patch as part of the build.
Signed-off-by: Boaz Sade <boaz@dragonflydb.io>

### DIFF
--- a/patches/mimalloc-v2.0.5.patch
+++ b/patches/mimalloc-v2.0.5.patch
@@ -1,0 +1,73 @@
+From 41d8a6b4995a87b2c5e41c8a59bb9a326bca2755 Mon Sep 17 00:00:00 2001
+From: Boaz Sade <boaz@dragonflydb.io>
+Date: Tue, 22 Nov 2022 16:35:47 +0200
+Subject: [PATCH] code chagnes for the df
+
+---
+ include/mimalloc-types.h |  6 +++---
+ include/mimalloc.h       |  2 ++
+ src/alloc.c              | 21 +++++++++++++++++++++
+ 3 files changed, 26 insertions(+), 3 deletions(-)
+
+diff --git a/include/mimalloc-types.h b/include/mimalloc-types.h
+index 310fb92..d0b5073 100644
+--- a/include/mimalloc-types.h
++++ b/include/mimalloc-types.h
+@@ -63,9 +63,9 @@ terms of the MIT license. A copy of the license can be found in the file
+ 
+ // Encoded free lists allow detection of corrupted free lists
+ // and can detect buffer overflows, modify after free, and double `free`s.
+-#if (MI_SECURE>=3 || MI_DEBUG>=1 || MI_PADDING > 0)
+-#define MI_ENCODE_FREELIST  1
+-#endif
++//#if (MI_SECURE>=3 || MI_DEBUG>=1 || MI_PADDING > 0)
++//#define MI_ENCODE_FREELIST  1
++//#endif
+ 
+ 
+ // ------------------------------------------------------
+diff --git a/include/mimalloc.h b/include/mimalloc.h
+index 83debd2..90d6ae2 100644
+--- a/include/mimalloc.h
++++ b/include/mimalloc.h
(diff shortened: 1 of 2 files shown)


patches/mimalloc-v2.0.5.patch
```diff
@@ -0,0 +1,73 @@
...
+
+diff --git a/include/mimalloc-types.h b/include/mimalloc-types.h
+index 310fb92..d0b5073 100644
+--- a/include/mimalloc-types.h
++++ b/include/mimalloc-types.h
+@@ -63,9 +63,9 @@ terms of the MIT license. A copy of the license can be found in the file
+ 
+ // Encoded free lists allow detection of corrupted free lists
+ // and can detect buffer overflows, modify after free, and double `free`s.
+-#if (MI_SECURE>=3 || MI_DEBUG>=1 || MI_PADDING > 0)
```
romange, line 20: what's the reason for commenting out these defines?

boazsade: When this is comment out, it would make the struct mi_page_t the same size for both debug and release version. I then tested this change, and it was working. At first I though that I would just remove the extra parameter that is added for the debug version, but it was reference in many places as well as function that are activated when this flag is on. So I though that the best option is just to deactivate it.

patches/mimalloc-v2.0.5.patch
```diff
@@ -0,0 +1,73 @@
...
+index 8cf7242..c2e8d0c 100644
+--- a/src/alloc.c
++++ b/src/alloc.c
+@@ -912,3 +912,24 @@ void* mi_new_reallocn(void* p, size_t newcount, size_t size) {
+     return mi_new_realloc(p, total);
+   }
+ }
++
++bool mi_heap_page_is_underutilized(mi_heap_t* heap, void* p, float ratio) mi_attr_noexcept {
++  mi_page_t* page = _mi_ptr_page(p);	// get the page that this belongs to
```
boazsade, line 52: I copy the code from the issue, I just change the first part to use a builtin function that return the page when a pointer is pass to it.

patches/mimalloc-v2.0.5.patch
romange: what is the size now when it's commented out?

note: we add `mi_heap_page_is_underutilized` here because we came to the conclusion that using structures from malloc-types is not robust and dangerous. if you fix this then maybe we can leave the code (function) in our repo...

boazsade: The size of the mi_page_t is now 80 for release (as it was before), and it is down from 96 to 80 in debug mode. As for the patch - I think it would be better this way, as we are less depending on getting code from their internals out into us. As you can see the function now is very short, as I have access to the function that "pull" the page based on pointer. I left the function "zmalloc_page_is_underutilized" in zmalloc.h as an entry point that only call the function in mimalloc. This way we don't need to include the mimallc-types.h header.

dranikpg: Just out of curiosity, why do we need it to be fixed size?

I understand that we use the mimalloc heap in very simple ways, but we give up on a lot of debug info.

romange: we usually do not use these advanced features anyway because we always compile mimalloc in release mode.
What Boaz tries to do is to allow us to include malloc-types from dragonfly code and access its internal data structures.

it's not possible to do when the same header file is compiled with different defines from mimalloc project. 
I do not feel strongly about it - we decided to add the function to mimalloc anyways.

dranikpg: Hm, I get it. 

> when we run in debug mode and mimalloc is compiled in release

In this case you could just compile mimacalloc with the same settings and #ifdef in df code

dranikpg (approved):

romange (approved):

PR #29, opened 2022-12-10
BlockingCounter Cancel + small utils

dranikpg:


base/io_buf.h
```diff
@@ -65,6 +65,15 @@ class IoBuf {
     size_ += sz;
   }
```
dranikpg, line 67: By the way, there is no reason for IoBuf to be immovable, but this requires implementing the move constructor manually because we can't have a default-default constructor (as the constructor with default value capacity=256) conflicts with it

romange: Have not put too much thought.into it. Feel free to extend if it blocks you in Dragonfly

romange (approved):

PR #34, opened 2022-12-20
support for https added as well as an example application

boazsade:
This add support for HTTPS client.
The code is based on the existing HTTP client and the TLS socket. What it adds is:
* HTTPS client that uses the TLS socket under the hood.
* A demo application that can be use to demo this new client.
note that the application is very useful to ensure that the TLS is working successfully.
I tried to make sure that this will be as secure as possible without making the use of it to complicated (i.e. loading cert file for example).
Signed-off-by: Boaz Sade <boaz@dragonflydb.io>

### DIFF
--- a/examples/https_client_cli.cc
+++ b/examples/https_client_cli.cc
@@ -1,0 +1,186 @@
+// Create an https client test program
+// Note this that example show how to run HTTPS client
+// which connect to the remote server, read a resource and print thi
+// the response to the LOG.
+// To see command line args run with --helpful command line flag
+// Since there is no reasonable default host name you must at minimum run
+// this with --host <domain name: https://en.wikipedia.org/wiki/Domain_name> such as www.google.com
+// (note that you should not add the service name at the start such as https).
+// To demo that this is not a blocking HTTPS client this would run 2 fibers for connection
+// and another task that basically does nothing.
+// output from this is either an error (in case we failed - try given illegal host or port) or a
+// the headers and the body from the host and resource you gave.
+// for example:
+// ./https_client --host redis.io --resource /commands/asking/ --vmodule=https_client=2
+// --logbuflevel=-1 --alsologtostderr
+// will print the content from https://redis.io/commands/asking command to the log as well as to the
+// terminal
+#include <openssl/err.h>
+
+#include <boost/beast/http.hpp>
+#include <chrono>
+#include <iostream>
+#include <memory>
+
+#include "base/flags.h"
+#include "base/init.h"
+#include "util/fibers/event_count.h"
+#include "util/http/http_client.h"
+#include "util/proactor_base.h"
+#include "util/proactor_pool.h"
+#include "util/tls/tls_engine.h"
+#include "util/tls/tls_socket.h"
+#include "util/uring/uring_pool.h"
+
+using util::ProactorBase;
+using util::ProactorPool;
+using util::http::Client;
+using util::http::TlsClient;
+namespace h2 = boost::beast::http;
+using ResponseType = h2::response<h2::string_body>;
+using absl::GetFlag;
+
+ABSL_FLAG(short, http_port, 443, "HTTPS remote server port.");
+ABSL_FLAG(std::string, host, "", "Remote host domain name");
+ABSL_FLAG(std::string, resource, "/", "resource path on the remote host");
+
+struct RequestResults {
+  unsigned int status = 400;
+  std::optional<ResponseType> result;
+
+  RequestResults() = default;
+
+  RequestResults(unsigned int s, ResponseType&& res) : status{s}, result{std::move(res)} {
+  }
+
+  bool good() const {
+    return status < 400;
+  }
+
+  bool has_result() const {
+    return result != std::nullopt;
+  }
+
+  const ResponseType& get() const {
+    assert(good() && has_result());
+    return result.value();
+  }
+};
+
+std::ostream& operator<<(std::ostream& os, const RequestResults& res) {
+  if (res.has_result()) {
+    if (res.good()) {
+      const auto& msg = res.get();
+      os << "headers: " << msg.base();
+      return os << "body: " << msg.body();
+    } else {
+      return os << "got errors results: " << res.result.value();
+    }
+  } else {
+    return os << "no result were successfully read from remote host";
+  }
+}
+
+using OpResult = io::Result<RequestResults, std::string>;
+
+OpResult ConnectAndRead(ProactorBase* proactor, std::string_view host, std::string_view service,
+                        const char* resource, SSL_CTX* ssl_ctx) {
+  TlsClient http_client{proactor};
+
+  http_client.set_connect_timeout_ms(2000);
+  auto list_res = proactor->Await([&] {
+    if (auto ec = http_client.Connect(host, service, ssl_ctx); !ec) {
+      h2::request<h2::string_body> req{h2::verb::get, resource, 11 /*http 1.1*/};
+      req.set(h2::field::host, std::string(host).c_str());
+      ResponseType res;
+      if (auto ec = http_client.Send(req, &res); !ec) {
+        return OpResult{RequestResults{res.result_int(), std::move(res)}};
+      } else {
+        return OpResult{
+            nonstd::unexpected<std::string>("Failed HTTPS GET Send(): " + ec.message())};
+      }
+    } else {
+      return OpResult{nonstd::unexpected<std::string>("Failed to connect: " + ec.message())};
+    }
+    return OpResult{RequestResults{}};
+  });
+
+  return list_res;
+};
+
+using namespace std::chrono_literals;
+
+util::fibers_ext::Fiber ReadFromRemoteHost(ProactorPool* pool, SSL_CTX* ssl_context,
+                                           std::string_view task_name,
+                                           util::fibers_ext::Done* done_condition) {
+  std::string service = std::to_string(GetFlag(FLAGS_http_port));
+  std::string host = GetFlag(FLAGS_host);
+  std::string resource = GetFlag(FLAGS_resource);
+
+  LOG(INFO) << task_name << ": host '" << host << "', port " << service << ", resource '"
+            << resource << "'";
+  return pool->GetNextProactor()->LaunchFiber([=]() {
+    while (true) {
+      if (done_condition->WaitFor(1s)) {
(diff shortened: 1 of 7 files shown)


examples/https_client/https_client.cc
```diff
@@ -0,0 +1,184 @@
...
+using util::ProactorBase;
+using util::ProactorPool;
+using util::http::Client;
+using util::http::SecureClient;
+namespace h2 = boost::beast::http;
+using ResponseType = h2::response<h2::string_body>;
+using absl::GetFlag;
+
+ABSL_FLAG(std::string, http_port, "443", "HTTPS remote server port.");
+ABSL_FLAG(std::string, host, "", "Remote host domain name;
```
romange, line 44: did you try to compile it?

examples/https_client/https_client.cc
```diff
@@ -0,0 +1,184 @@
+// Create an https client test program
```
romange, line 1: please move it to examples folder and rename it to https_client_cli

examples/https_client/https_client.cc
boazsade: sure - I was using this all the time to verify that its working

util/tls/tls_socket.h
```diff
@@ -7,12 +7,12 @@
...
 
 #include "util/fiber_socket_base.h"
-
+#include "util/tls/tls_engine.h"
 
 namespace util {
 namespace tls {
 
-class Engine;
+// class Engine;
```
romange, line 15: ?

examples/https_client/https_client.cc
```diff
@@ -0,0 +1,184 @@
...
+    LOG(ERROR) << "Host name is missing - please run with --host <host name>";
+    return -1;
+  }
+
+  SSL_library_init();
+  SSL_load_error_strings();
+  util::uring::UringPool pp;
+  int count = 0;
+  pp.Run();
+  SSL_CTX* ssl_context = SecureClient::CreateSslContext(SecureClient::LEVEL_2);
```
romange, line 168: who frees this context?

util/http/http_client.cc
```diff
@@ -97,5 +108,47 @@ bool Client::IsConnected() const {
   return socket_ && socket_->IsOpen();
 }
 
+///////////////////////////////////////////////////////////////////////////////
+
+SSL_CTX* SecureClient::CreateSslContext(SecurityLevel sec_level) {
+  SSL_CTX* ctx = SSL_CTX_new(TLS_client_method());
+  if (ctx) {
+    long tls_ver = sec_level == LEVEL_5 ? TLS1_3_VERSION : TLS1_2_VERSION;
```
romange, line 116: why not use tls1.3 hardcoded?

util/tls/tls_socket.h
boazsade: I had to include the file with the Engine class because of the "Handshake" function that I've added here

examples/https_client/https_client.cc
boazsade: You're right - I will add a "free" at the end

util/http/http_client.cc
```diff
@@ -97,5 +108,47 @@ bool Client::IsConnected() const {
...
+    SSL_CTX_set_min_proto_version(ctx, tls_ver);
+    if (SSL_CTX_set_default_verify_paths(ctx) != 1) {
+      LOG(WARNING) << "failed to set default verify path on client context for TLS connection";
+    }
+    SSL_CTX_set_options(ctx, SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS);
+    SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, VerifyCallback);
+    SSL_CTX_set_verify_depth(ctx, 4);
+    // Can also be defined using "ADH:@SECLEVEL=0" cipher string below.
+    SSL_CTX_set_security_level(ctx, int(sec_level));
```
romange, line 125: what happens if you do not set it? i.e. why default is not good?

util/http/http_client.cc
boazsade: We can, it use to be the default from the example I took to use TLS version 1.2, but I will change

util/http/http_client.cc
```diff
@@ -97,5 +108,47 @@ bool Client::IsConnected() const {
...
+    SSL_CTX_dane_enable(ctx);  // see https://www.internetsociety.org/resources/deploy360/dane/
+    DCHECK(1 == SSL_CTX_set_cipher_list(ctx, "ADH"));
```
romange, line 127: why ADH is important here? specifically it's confusing because this call won't happen in the release mode

util/http/http_client.h
```diff
@@ -128,5 +133,34 @@ template <typename Req> std::error_code Client::Send(const Req& req) {
   return HandleError(ec);
 }
 
+///////////////////////////////////////////////////////////////////////////////
+
+class SecureClient : public Client {
```
romange, line 138: maybe call it TlsClient or SslClient?

util/http/http_client.cc
boazsade: The issue with default is that I don't know what it uses. In any case, when I tried to use level 3 and above the verification would failed (for the TLS handshake) I can assume that it was because the key was not long enough. From the example I took it was set to 0 which is not good enough.

util/http/http_client.cc
boazsade: I will remove it

util/http/http_client.h
boazsade: OK (Tls is better)

util/http/http_client.h
```diff
@@ -128,5 +133,34 @@ template <typename Req> std::error_code Client::Send(const Req& req) {
...
+  // out of the box
+  static SSL_CTX* CreateSslContext(SecurityLevel sec_level);
+
+  explicit SecureClient(ProactorBase* proactor) : Client(proactor) {
+  }
+
+  std::error_code Connect(StringPiece host, StringPiece service, SSL_CTX* context);
+
+ private:
+  using SecureSocket = util::tls::TlsSocket;
```
romange, line 160: you do not use SecureSocket here. I prefer that you just declare `using util::tls::TlsSocket` in cc than introducing another alias.

util/http/http_client.h
```diff
@@ -128,5 +133,34 @@ template <typename Req> std::error_code Client::Send(const Req& req) {
...
+
+  std::unique_ptr<FiberSocketBase> connected_socket_;
```
romange, line 162: maybe call it `tcp_socket_` ?

util/http/http_client.h
```diff
@@ -128,5 +133,34 @@ template <typename Req> std::error_code Client::Send(const Req& req) {
   return HandleError(ec);
 }
 
+///////////////////////////////////////////////////////////////////////////////
+
+class SecureClient : public Client {
+ public:
+  // see https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_security_level.html
+  enum SecurityLevel {
```
romange, line 141: I prefer avoiding to duplicate APIs for this. If sane defaults work then we do not need it.
if they do not work then a caller that receives SSL_CTX object cal call whatever openssl function he desires.

util/tls/tls_socket.h
romange: i understood, so why not delete the line?

util/tls/tls_socket.h
```diff
@@ -43,6 +43,13 @@ class TlsSocket : public FiberSocketBase {
 
   SSL* ssl_handle();
 
+  // Note:
+  // This is not part of the FiberSocketBase.
+  // This function is mostly useful for client side,
+  // since part of the TLS is to handshake after the connection
+  // is done using "normal" TCP connection.
+  error_code Handshake(Engine::HandshakeType Handshake_type);
```
romange, line 51: who do you need to expose this method? why not put this functionality inside TlsSocket::Connect method that is currently empty?

util/tls/tls_socket.h
romange: Also, see my comment below. I do not think handshake must be exposed.

examples/https_client/https_client.cc
boazsade: I see what the problem, I did some cleanup before the push and remove the end of the flag, I'll fix it

romange: it was a retorical question :)
This version doesn't  compile. You can see it on the screen, it lacks parenthesis.

romange: and you can see the color of the text below. it should raise red flags.

util/http/http_client.h
boazsade: OK

util/http/http_client.h
boazsade: OK, I didn't like the name I come up with :)

util/tls/tls_socket.h
boazsade: I tried put it here (it was there originally, don't remember why I remove it from there, I will try to put it back here again, maybe because of other issues I had I though I should move it out

util/tls/tls_socket.h
boazsade: removed

util/http/http_client.h
boazsade: OK, I will remove it

examples/https_client/https_client.cc
boazsade: moved

util/http/http_client.h
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <openssl/ssl.h>  // required by SSL_CTX
```
romange, line 7: why do you need ssl in this file?

examples/https_client_cli.cc
```diff
@@ -0,0 +1,186 @@
...
+  // Let the background tasks run for a while
+  for (int i = 0; i < 3; i++) {
+    util::fibers_ext::SleepFor(1s);
+  }
+  done_condition.Notify();
+  fiber.Join();
+  fiber2.Join();
+  idle_fiber.Join();
+  LOG(INFO) << "finish running the test";
+  TlsClient::FreeContext(ssl_context);
```
romange, line 185: I do not see you added these files to the PR.

romange (approved):

PR #37, opened 2022-12-30
Cleanup IoBuf and add source/sink for it

dranikpg:
Some things that I desperately missed in DF:
- IoBuf is not movable (And is wrapped often into unique_ptr to account for this 🤯 )
- IoBuf WriteAndCommit akin to ReadAndConsume
- IoBuf cannot be used as a sink or as a source
- There is no way to convert Bytes into string_view easily

### DIFF
--- a/base/io_buf.h
+++ b/base/io_buf.h
@@ -79,7 +79,7 @@ class IoBuf {
 
   // Copy num_copy bytes from source to append buffer and mark them as written.
   // Ensures append buffer is large enough.
-  void WriteAndCommit(size_t num_copy, const void* source);
+  void WriteAndCommit(const void* source, size_t num_copy);
 
   // Ensure required append buffer size.
   void EnsureCapacity(size_t sz) {
--- a/base/io_buf.h
+++ b/base/io_buf.h
@@ -10,17 +10,16 @@
 
 namespace base {
 
+// Generic buffer for reads and writes.
+// Write directly to AppendBuffer() and mark bytes as written with CommitWrite.
+// Read from InputBuffer() and mark bytes as read with ConsumeInput.
 class IoBuf {
  public:
-  IoBuf(const IoBuf&) = delete;
-  IoBuf& operator=(const IoBuf&) = delete;
-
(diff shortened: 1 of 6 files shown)


base/io_buf.h
```diff
@@ -44,85 +56,61 @@ class IoBuf {
...
-      }
-      delete[] buf_;
-    }
-
-    buf_ = nb;
-    capacity_ = sz;
-  }
+  // Reserve by whole buffer size.
+  // Use EnsureCapacity instead for resizing only by desired append buffer size.
+  void Reserve(size_t full_size);
```
dranikpg, line 91: Those function is dangerously misleading, it accounts only for the TOTAL size. Most of the time EnsureCapacity is really needed from outside accessors

io/io.cc
```diff
@@ -73,6 +73,22 @@ void AsyncWriteState::OnCb(Result<size_t> res) {
   owner->AsyncWriteSome(cur, arr.end() - cur, [this](Result<size_t> res) { this->OnCb(res); });
 }
 
+// Read some bytes from buf starting at given offset offs and update it.
+Result<size_t> ReadSomeBytes(const iovec* v, uint32_t len, Bytes buf, off_t* offs) {
```
romange, line 77: please rename v into dest, it's hard to understand from the signature what's the input and what's the output.

base/io_buf.h
```diff
@@ -44,85 +56,61 @@ class IoBuf {
...
+    size_ += num_written;
   }
 
-  void Clear() {
-    size_ = 0;
-    offs_ = 0;
-  }
+  // Copy num_copy bytes from source to append buffer and mark them as written.
+  // Ensures append buffer is large enough.
+  void WriteAndCommit(size_t num_copy, const void* source);
```
romange, line 82: usually a ptr is followed by length

base/io_buf.h
romange: if you think it's not needed, feel free to remove it...

romange (approved):

PR #48, opened 2023-03-11
string_view_sso will use 22 bytes SSO when compiled with libc++

chakaz:
This will make it compatible with libc++'s std::string.

Alternatively, we could always use 22 bytes (even for STL implementations supporting smaller SSOs, such as libstdc++).


romange:
Lgtm. I do not even know what libc++ is.

romange (approved):

chakaz:
> Lgtm. I do not even know what libc++ is.

:)
libc++ is a great STL implementation, built for (and by) clang. It's superior in many ways to libstdc++, which is the library used by GCC (and can also be used by clang).
There's some more info here if you're interested: https://shaharmike.com/cpp/std-string/

PR #54, opened 2023-03-23
Use c-ares to resolve DNS requests.

chakaz:
This PR is related to dragonfly's issue: https://github.com/dragonflydb/dragonfly/issues/947

### DIFF
--- a/util/dns_resolve.cc
+++ b/util/dns_resolve.cc
@@ -4,10 +4,12 @@
 
 #include "util/dns_resolve.h"
 
+#include <ares.h>
 #include <arpa/inet.h>
 #include <netdb.h>
 #include <signal.h>
 #include <string.h>
+#include <sys/epoll.h>
 #include <sys/socket.h>
 #include <sys/types.h>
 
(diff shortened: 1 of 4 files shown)


util/dns_resolve.cc
```diff
@@ -4,126 +4,83 @@
...
+  ares_channel channel;
+  CHECK_EQ(ares_init(&channel), ARES_SUCCESS);
+
+  DnsResolveCallbackArgs cb_args;
+  cb_args.dest_ip = dest_ip;
+  boost::fibers::future<std::error_code> future = cb_args.promise.get_future();
+  ares_gethostbyname(channel, dns, AF_INET, &DnsResolveCallback, &cb_args);
 
-  freeaddrinfo(servinfo);
+  WaitForChannel(channel);
```
romange, line 80: You are constantly polling here. Is it how they advice to use c-ares? 
Am i wrong claiming that the future is already resolved when this function returns?

chakaz: > You are constantly polling here. Is it how they advice to use c-ares?
The way one should use `ares_process` is via polling, here's an example: https://c-ares.org/ares_process.html
Because we're using Fibers here, I figured that this is the way to go, as I was under the impression that Fibers do not play along with `select` / `poll` (?), and so every time there's a slot to run our Fiber it'll quickly process the ares request.
*However*, re-reading the example and comparing with the original PR, I realized that it was not ideal in terms of yielding (it might've called `yield()` when it was done processing, or call `ares_process()` when no FD was ready). I slightly modified the code to fix that, PTAL.

> Am i wrong claiming that the future is already resolved when this function returns?
You are not wrong. By the time the function returns, the future is already resolved. Is this an issue?

romange: Yes, your code technically works but if DNS resolution takes 0.5s then you are going to spin for 0.5s with 100% CPU utilization. It's not an end of the world of course. However, after skimping over docs, I understand now that doing something more sophisticated will require knowledge of iouring API and it's not productive at this point. so lets leave it as is.
Btw, see how chatgpt provided me a reference code that would probably help me integrate c-ares into helio native io/engine without using select and polling. But this is out of scope for this PR.

Regarding futures - yes, they are redundant the way you use them.  I will give concrete tips in the code.

util/dns_resolve.cc
```diff
@@ -4,126 +4,88 @@
...
-// TODO: The correct solution would be to use c-ares (see the comment below).
-// #define USE_GAI_ASYNC
-#endif
-
-#ifdef USE_GAI_ASYNC
-struct NotifyStruct {
-  atomic_bool gate{false};
-  fibers_ext::EventCount evc;
+struct DnsResolveCallbackArgs {
+  boost::fibers::promise<std::error_code> promise;
```
romange, line 27: just use `std::error_code` without promise here.

util/dns_resolve.cc
```diff
@@ -4,126 +4,88 @@
...
-  delete ns;
-
-  int res = gai_error(&gai_req);
-  if (res == EAI_CANCELED) {
-    return make_error_code(errc::operation_canceled);
+void DnsResolveCallback(void* ares_arg, int status, int timeouts, hostent* hostent) {
+  CHECK(ares_arg != nullptr);
+  auto* cb_args = static_cast<DnsResolveCallbackArgs*>(ares_arg);
+  if (status != ARES_SUCCESS || hostent == nullptr) {
+    cb_args->promise.set_value(make_error_code(errc::address_not_available));
```
romange, line 35: just assign to error_code, no need for promise

util/dns_resolve.cc
```diff
@@ -4,126 +4,88 @@
...
-    return make_error_code(errc::address_not_available);
+  char** addr = hostent->h_addr_list;
+  if (addr == nullptr) {
+    cb_args->promise.set_value(make_error_code(errc::address_not_available));
+    return;
   }
 
-#endif
+  ares_inet_ntop(AF_INET, *addr, cb_args->dest_ip, INET_ADDRSTRLEN);
+  cb_args->promise.set_value(error_code());
```
romange, line 51: redundant line

util/dns_resolve.cc
```diff
@@ -4,126 +4,88 @@
...
+
+}  // namespace
+
+error_code DnsResolve(const char* dns, uint32_t wait_ms, char dest_ip[]) {
+  ares_channel channel;
+  CHECK_EQ(ares_init(&channel), ARES_SUCCESS);
+
+  DnsResolveCallbackArgs cb_args;
+  cb_args.dest_ip = dest_ip;
+  boost::fibers::future<std::error_code> future = cb_args.promise.get_future();
```
romange, line 82: no need, just return `cb_args.result` at the end. Since you have this barrier `WaitForChannel` you already take care of resolution, and this is why using futures is redundant here

util/dns_resolve.cc
```diff
@@ -4,126 +4,88 @@
...
+#include <ares.h>
 #include <arpa/inet.h>
 #include <netdb.h>
 #include <signal.h>
 #include <string.h>
 #include <sys/socket.h>
 #include <sys/types.h>
 
+#include <atomic>
+#include <boost/fiber/future.hpp>
```
romange, line 16: no need

util/dns_resolve.cc
romange: actually chatgpt4 produced a better version with lambdas and everything 😆 
```cpp
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/epoll.h>
#include <ares.h>
#include <arpa/inet.h>
#include <netinet/in.h>
#include <unistd.h>

static void callback(void *arg, int status, int timeouts, struct hostent *hostent) {
    if (status == ARES_SUCCESS) {
        char ip[INET_ADDRSTRLEN];
        inet_ntop(hostent->h_addrtype, hostent->h_addr_list[0], ip, sizeof(ip));
        printf("Resolved IP: %s\n", ip);
    } else {
        printf("Failed to resolve: %s\n", ares_strerror(status));
    }
}

int main(int argc, char *argv[]) {
    if (argc != 2) {
        printf("Usage: %s <hostname>\n", argv[0]);
        return 1;
    }

    int epoll_fd = epoll_create1(0);
    if (epoll_fd < 0) {
        perror("epoll_create1");
        return 1;
    }

    ares_library_init(ARES_LIB_INIT_ALL);

    struct ares_options options;
    options.sock_state_cb_data = (void *)(intptr_t)epoll_fd;
    options.sock_state_cb = [](void *data, ares_socket_t sock, int read, int write) {
        int epoll_fd = (intptr_t)data;

        struct epoll_event ev;
        ev.data.fd = sock;
        ev.events = 0;
        if (read)
            ev.events |= EPOLLIN;
        if (write)
            ev.events |= EPOLLOUT;

        if (epoll_ctl(epoll_fd, EPOLL_CTL_MOD, sock, &ev) < 0) {
            perror("epoll_ctl");
        }
    };

    ares_channel channel;
    if (ares_init_options(&channel, &options, ARES_OPT_FLAGS | ARES_OPT_SOCK_STATE_CB | ARES_OPT_SOCK_STATE_CB_DATA) != ARES_SUCCESS) {
        fprintf(stderr, "ares_init_options failed\n");
        ares_library_cleanup();
        return 1;
    }

    ares_gethostbyname(channel, argv[1], AF_INET, callback, NULL);

    struct epoll_event events[16];
    int nfds;
    while (1) {
        nfds = epoll_wait(epoll_fd, events, 16, -1);
        if (nfds < 0) {
            perror("epoll_wait");
            break;
        }

        for (int i = 0; i < nfds; ++i) {
            ares_socket_t sock = events[i].data.fd;
            int read_ready = (events[i].events & EPOLLIN) != 0;
            int write_ready = (events[i].events & EPOLLOUT) != 0;

            ares_process_fd(channel, read_ready ? sock : ARES_SOCKET_BAD, write_ready ? sock : ARES_SOCKET_BAD);
        }

        if (ares_fds(channel, NULL, NULL) == 0) {
            break;
        }
    }

    ares_destroy(channel);
    ares_library_cleanup();
    close(epoll_fd);
    return 0;
}
```

chakaz: Cool code by chatgpt / chatgpt4. I think I spotted some errors there, but that's actually useful. I should get used to asking them questions actually.

Re/ future: yeah, that's definitely leftover from before I added the wait function I removed them per your review.

chakaz: Oh, and regarding live wait.

My understanding of Fibers is that if the current Fiber `yield()`s (like I do when the FDs are not ready), then there's a context switch to another Fiber waiting to run. In other words, per my understanding, this code will only consume 100% CPU in the case where no other Fiber (on the same thread) needs to do any work. That doesn't sound very concerning, right? Or am I misunderstanding something about the Fibers mechanism (or my code)?

romange: It's true that yield() passes control to other fibers. However, a typical cycle for any reactive system is to handle pending cpu work and then, when nothing is left to process - to block on  I/O (epoll_wait). This is why servers are not constantly at 100% CPU and this is how you can use monitoring to actually measure the effective load on your servers. I am not talking even talking about energy waste. So yes - the code works, but no, it is not production ready because we would not want random spikes of 100% cpu usage that will send alerts and confuse devops.

romange: I know how to "plug" it into a scheduler but I do not know how to do it with c-ares. I suspect we need to use lower-level c-ares API, something like what gpt4 showed me.

Regarding the scheduler - basically we need to subscribe to io events (read and write) by passing the waking callback to the I/O engine and then suspend ourselves.  This should replace select + yield.

The reality is a bit more complicated because we support 2 i/o engines - epoll based and iouring based.
Lets start with iouring, because it's our default engine. You can look at `util::uring::FiberCall` which does exactly that - suspending and waking. You can look at uring_socket.cc on how it's used, but basically you need to 
define a FiberCall variable and then use `->` to access SubmitEntry interface. I am guessing you will need to use `PrepPollAdd` method to subscribe for your events.  Finally, `FiberCall::Get` suspends and awakes when an event arrives and then you can continue calling c-ares.

However, if I were you, I would start with converting gpt4 code (or any other example) and creating a simple C-like program that works with c-ares and epoll_wait. You can even add it under examples. Once you figure out how to create a regular program, adapting it to helio fibers will be much easier.

chakaz: Gotcha. Sounds good. Please see the latest commit, using `epoll_wait()`.

romange:
Looks like a step in the right direction!

Can you please squash all changes and ask to merge into a branch instead of master?

chakaz:
Could it be the case that you'll have to create the branch for me?
I can't find a way to do so (and I guess it makes sense for random people not to have such access?)

romange:
sorry, https://github.com/romange/helio/tree/Cares created

chakaz:
Ok, I think I did that. Still new to git, lmk if something's wrong.

chakaz:
@romange Following your above guidelines, I'm trying to look at how to use the new code with the 2 i/o engines:
1. How can I tell which I/O engine is active at runtime?
2. It looks like it would be easier to separate the actual implementation (at least the part that interacts with the I/O engine) to separate files, as I'm having some hard time including both engines in the same file. Would you agree?

romange:
Oh man, helio is mess right now. I am trying to switch to FB2 implementation for fibers so all the new code is under "util/fibers" under library fibers2.

Currently, only EpollProactor (partly) supports the interface you need. Yes, it's a mess but solvable. 
I just submitted a PR that allows running echo_server with a not completed version of DnsResolve.
you can run it  with `./echo_server --epoll --connect foobar.com --logtostderr --vmodule=dns_resolve=1`
Please note it always fails right now because it's WIP.  You should change your build config with `ccmake .` , switch "USE_FB2" to ON and reconfigure it.
What's left to implement, imho is `UpdateSocketsCallback` using the fb2::EpollProactor interface.
I can do it, frankly because I do not know how much joy you will get from working within a temporary mess of what helio is right now.

romange:
And then of course, after it's done with Epoll version, I would not lift the abstract interface to fb2::ProactorBase, and implement it with Uring implementation as well. Unfortunately, it's not gonna be a totally clean code in DnsResolve because iouring does not support multi-shot EPOLL registrations (well, supports only in the latest kernel versions so I do not want to use it).

PR #56, opened 2023-03-29
Save SWAP information from `/proc/meminfo`

talbii:
Save SWAP information from `/proc/meminfo`, can also be used in determining whether the SWAP file is enabled or not.

### DIFF
--- a/io/proc_reader.h
+++ b/io/proc_reader.h
@@ -23,6 +23,9 @@ struct MemInfoData {
   size_t mem_buffers = 0;
   size_t mem_cached = 0;
   size_t mem_SReclaimable = 0;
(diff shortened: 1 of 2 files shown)


io/proc_reader.h
```diff
@@ -23,6 +23,9 @@ struct MemInfoData {
   size_t mem_buffers = 0;
   size_t mem_cached = 0;
   size_t mem_SReclaimable = 0;
+  size_t swap_cached = 0;
```
romange, line 26: Weird indentation

romange: Weird, will take look later

romange (approved):

PR #58, opened 2023-03-31
allow to compile with an OpenSSL installation

vvhungy:
- remove the override of OPENSSL_ROOT_DIR
- allow to pass a custom OpenSSL folder from build command as -DOPENSSL_ROOT_DIR=/path/to/openssl/folder
- make sure crypto and ssl library using the same installation source


romange (approved):

romange:
Thanks!

PR #63, opened 2023-04-10
fix: Poll socket if connect() returns EINPROGRESS

royjacobson:


romange:
This should not happen because iouring should know how to handle
einprogress internally.

1. Are you sure it happens with uring socket and not epoll?
2. How do you reproduce this?

royjacobson:
> This should not happen because iouring should know how to handle einprogress internally. 1. Are you sure it happens with uring socket and not epoll? 2. How do you reproduce this?

Apparently this can happen if the socket has a timeout. I based this code on the way it is done [in liburing](https://github.com/axboe/liburing/blob/4fed79510a189cc7997f6d04855ebf7fb66cc323/test/connect.c#L152)

I didn't manage to reproduce this in a unit test, `replication_test.py` in current dragonflydb master reproduces it very reliably.

royjacobson:
seems to have been caused by the same root issue as the one fixed in https://github.com/romange/helio/commit/7322b320c15baf6ac504d897d0a7f5b440510248

PR #70, opened 2023-05-05
Fix ownership issues when notifying fibers

romange:
Specifically, move ActivateOther under lock because in case a remote fiber times out, its liveness is not guaranteed outside of the lock.

### DIFF
--- a/util/fibers/synchronization.cc
+++ b/util/fibers/synchronization.cc
@@ -55,7 +55,6 @@ void Mutex::unlock() {
 
   detail::FiberInterface* fi = &wait_queue_.front();
   wait_queue_.pop_front();
-  lk.unlock();
   active->ActivateOther(fi);
 }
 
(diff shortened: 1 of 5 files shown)


romange:
The tests reproduce the crash, but remote fiber life span issue is not something I can reproduce. However, I believe the bug still exists. ActivateOther is lock-free (even wait-free) so I believe it should not add much to spinlock contention.

util/fibers/synchronization.cc
```diff
@@ -65,20 +64,20 @@ void CondVarAny::notify_one() noexcept {
...
 }
 
 void CondVarAny::notify_all() noexcept {
   decltype(wait_queue_) tmp_queue;
+  detail::FiberInterface* active = detail::FiberActive();
+
+  unique_lock lk(wait_queue_splk_);
   wait_queue_splk_.lock();
   tmp_queue.swap(wait_queue_);
-  wait_queue_splk_.unlock();
```
dranikpg, line 79: Seems like you forgot to remove the explicit lock()

dranikpg (approved):

PR #78, opened 2023-05-22
feat: Add a ergonomic function for permissive join

royjacobson:


util/fibers/fiber2.h
```diff
@@ -70,6 +70,9 @@ class Fiber {
 
   void Join();
 
+  // Join fiber if it's running, else do nothing.
+  void PermJoin();
```
chakaz, line 74: Reading this, I wouldn't know `Perm` stands for "permissive" (I woulda thought about permanent maybe?)
How about `JoinIfJoinable()`?
Anyway, looks like a good utility :)

chakaz (approved):

util/fibers/fiber2.h
royjacobson:
```suggestion
  void JoinIfNeeded();
```

util/uring/uring_socket_test.cc
```diff
@@ -85,9 +85,7 @@ void UringSocketTest::TearDown() {
 
   listen_socket_.reset();
   conn_socket_.reset();
-  if (accept_fb_.IsJoinable()) {
-    accept_fb_.Join();
-  }
+  accept_fb_.PermJoin();
```
royjacobson, line 88:
```suggestion
  accept_fb_.JoinIfNeeded();
```

PR #86, opened 2023-06-21
Build with clang and `-Werror`

chakaz:


util/fibers/uring_proactor.h
```diff
@@ -129,7 +129,7 @@ class UringProactor : public ProactorBase {
   uint8_t sqpoll_f_ : 1;
   uint8_t register_fd_ : 1;
   uint8_t msgring_f_ : 1;
-  uint8_t reserved_f_ : 5;
+  [[maybe_unused]] uint8_t reserved_f_ : 5;
```
royjacobson: Can you just make it anonymous?

```suggestion
  uint8_t : 5;
```

royjacobson (commented):
I'm a bit worried that logging all those socket Close errors in WARNING has a high volume and high S/N ratio, but I'm not sure because I don't remember when we get an error on Close

util/fibers/uring_proactor.h
chakaz: That's a fantastic idea!

chakaz:
> I'm a bit worried that logging all those socket Close errors in WARNING has a high volume and high S/N ratio, but I'm not sure because I don't remember when we get an error on Close

I imagined it would be rare, but it's indeed a behavior change. I can just cast to `void` like I do in the tests, do you think that would be better?

royjacobson:
> > I'm a bit worried that logging all those socket Close errors in WARNING has a high volume and high S/N ratio, but I'm not sure because I don't remember when we get an error on Close
> 
> I imagined it would be rare, but it's indeed a behavior change. I can just cast to `void` like I do in the tests, do you think that would be better?

Ok, went to read the man page so I'm a bit wiser now :mage: 
Since the caller should only use the return code for diagnostics and the file descriptor is always closed - I say let's remove the `ABSL_MUST_USE_RESULT` from `Close()` in `fiber_socket_base.h`, and add `LOG(WARNING)` in the `Close()` method itself. From looking at possible errors it seems they should be rare enough that it's fine.

romange (approved):

chakaz:
@romange wdyt about changing `Shutdown()` and `Close()` to not return `void` rather than `std::error_code`, as proposed by @royjacobson ?
I'm happy to make this change, just making sure you'd like it as well before I put the time.

romange:
The problem with logging at low level is that you may get excessive logging
in some cases and you can not tune it since you do not control the context.
I am fine trying this approach and change it back if we learn it's too
noisy...

chakaz:
If you're uncertain, let's keep as is, at least for now

PR #87, opened 2023-06-23
feat(cmake): Adapt compile flags for s390x arch

iko1:
This PR comes after the compilation of adding s390x arch support in dragonflydb.

### DIFF
--- a/cmake/internal.cmake
+++ b/cmake/internal.cmake
@@ -105,6 +105,8 @@ if (NOT MARCH_OPT)
     set(MARCH_OPT "-march=sandybridge -mtune=skylake")
   elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
     # MacOS on arm64 - TBD.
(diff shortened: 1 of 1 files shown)


cmake/internal.cmake
```diff
@@ -105,6 +105,9 @@ if (NOT MARCH_OPT)
     set(MARCH_OPT "-march=sandybridge -mtune=skylake")
   elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
     # MacOS on arm64 - TBD.
+  elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "s390x")
+    set(MARCH_OPT "-march=native")
+    set(COMPILE_OPTS "${COMPILE_OPTS} -mzvector")
```
romange, line 110: you can add `-mzvector` MARCH_OPT, no need to add to `COMPILE_OPTS` imho

romange (approved):

PR #89, opened 2023-06-27
fix(dns): Support multiple sockets in parallel

chakaz:
This attempts to fix https://github.com/dragonflydb/dragonfly/issues/1486

### DIFF
--- a/util/fibers/dns_resolve.cc
+++ b/util/fibers/dns_resolve.cc
@@ -27,9 +27,9 @@ struct DnsResolveCallbackArgs {
 };
 
 struct AresSocketState {
-  bool has_reads = false;
-  bool has_writes = false;
+  unsigned mask;
   unsigned arm_index;
+  bool removed = false;  // Used to indicate that the socket was removed, without modifying the map
 };
 
 struct AresChannelState {
@@ -58,37 +58,10 @@ void UpdateSocketsCallback(void* arg, ares_socket_t socket_fd, int readable, int
   if (writable)
     mask |= ProactorBase::EPOLL_OUT;
 
-  auto [it, inserted] = state->sockets_state.try_emplace(socket_fd);
-  if (inserted) {
+  if (mask == 0) {
+    auto it = state->sockets_state.find(socket_fd);
+    CHECK(it != state->sockets_state.end());
     // TODO: to unify epoll management under a unified interface in ProactorBase.
-    if (state->proactor->GetKind() == ProactorBase::EPOLL) {
-      EpollProactor* epoll = (EpollProactor*)state->proactor;
-      auto cb = [state, socket_fd](uint32_t event_mask, int err, EpollProactor* me) {
-        AresSocketState& socket_state = state->sockets_state[socket_fd];
-        socket_state.has_writes = HasWrites(event_mask);
-        socket_state.has_reads = HasReads(event_mask);
-        if (state->fiber_ctx) {
-          detail::FiberActive()->ActivateOther(state->fiber_ctx);
-        }
-      };
-      it->second.arm_index = epoll->Arm(socket_fd, std::move(cb), mask);
-    } else {
-      CHECK_EQ(state->proactor->GetKind(), ProactorBase::IOURING);
-#ifdef __linux__
-      UringProactor* uring = (UringProactor*)state->proactor;
-      auto cb = [state, socket_fd](uint32_t event_mask) {
-        AresSocketState& socket_state = state->sockets_state[socket_fd];
-        socket_state.has_writes = HasWrites(event_mask);
-        socket_state.has_reads = HasReads(event_mask);
-        if (state->fiber_ctx) {
-          detail::FiberActive()->ActivateOther(state->fiber_ctx);
-        }
-      };
-      it->second.arm_index = uring->EpollAdd(socket_fd, std::move(cb), mask);
-#endif
-    }
-  } else {
-    CHECK_EQ(mask, 0u);
     if (state->proactor->GetKind() == ProactorBase::EPOLL) {
       EpollProactor* epoll = (EpollProactor*)state->proactor;
       epoll->Disarm(socket_fd, it->second.arm_index);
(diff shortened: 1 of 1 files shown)


romange (approved):

romange:
@chakaz  have you succeeded to reproduce it?

chakaz:
I'm afraid I was not able to repro it, but I do think that the proposed fix fixes a real bug. I do not think that c-ares guarantee to only use 1 socket, and from the assertion at https://github.com/dragonflydb/dragonfly/issues/1486#issue-1777202737 it certainly looks like 2 sockets are used (116 & 117).
I will continue trying to repro it if you think it's worth it

romange:
I agree there is a bug but I would like to be sure we indeed fix it. I suspect it's possible to reproduce the problem via docker environment. See, for example this SO: https://stackoverflow.com/questions/41717180/docker-compose-container-using-host-dns-server

it's pretty easy to run dragonfly via docker-compose (search our github) and maybe this way you can reproduce.

chakaz:
I was _finally_ able to reproduce this. It wasn't easy, and I end up setting a patched DNS server that responds with an error that causes c-ares to open another socket to retry.
Anyway, it's a good thing that you insisted on reproducing because my earlier commit had a bug that it assumed that `UpdateSocketsCallback()` must be called exactly once with each socket (this is _part_ of the same bug that exists prior to this PR).
Anyway, I fixed that too, and tested it both on a regular machine and also on an environment with the weird DNS setup.
PTAL.

chakaz:
BTW fwiw here's the stack that I was able to reproduce, which is identical to that of the user's:

```
> ./dragonfly --logtostderr --dbfilename=                                                                                                                                           
I20230629 06:53:57.509917  4227 init.cc:69] ./dragonfly running in opt mode.                                                                                                                                        
* Logs will be written to the first available of the following paths:                                                                                                                                               
/tmp/dragonfly.*                                                                                                                                                                                                    
./dragonfly.*                                                                                                                                                                                                       
* For the available flags type dragonfly [--help | --helpfull]                                                                                                                                                      
* Documentation can be found at: https://www.dragonflydb.io/docs                                                                                                                                                    
I20230629 06:53:57.510370  4227 dfly_main.cc:647] Starting dragonfly df-dev-0000000                                                                                                                                 
W20230629 06:53:57.510641  4227 dfly_main.cc:684] SWAP is enabled. Consider disabling it when running Dragonfly.                                                                                                    
I20230629 06:53:57.510715  4227 dfly_main.cc:687] maxmemory has not been specified. Deciding myself....                                                                                                             
I20230629 06:53:57.510782  4227 dfly_main.cc:691] Found 1.93GiB available memory. Setting maxmemory to 1.54GiB                                                                                                      
I20230629 06:53:57.511809  4228 uring_proactor.cc:158] IORing with 1024 entries, allocated 102720 bytes, cq_entries is 2048                                                                                         
I20230629 06:53:57.515367  4227 proactor_pool.cc:147] Running 1 io threads                                                                                                                                          
I20230629 06:53:57.519659  4227 server_family.cc:164] Data directory is "/home/shahar"                                                                                                                              
I20230629 06:53:57.522455  4228 listener_interface.cc:97] sock[7] AcceptServer - listening on port 6379                                                                                                             
F20230629 06:53:57.522627  4228 dns_resolve.cc:88] Check failed: state->channel_sock == socket_fd (8 vs. 9)                                                                                                         
*** Check failure stack trace: ***                                                                                                                                                                                  
    @     0x55b2ec5bcab0  google::LogMessage::Fail()                                                                                                                                                                
    @     0x55b2ec5c3f87  google::LogMessage::SendToLog()                                                                                                                                                           
    @     0x55b2ec5bc4c3  google::LogMessage::Flush()                                                                                                                                                               
    @     0x55b2ec5bde5f  google::LogMessageFatal::~LogMessageFatal()                                                                                                                                               
    @     0x55b2ec593397  util::fb2::(anonymous namespace)::UpdateSocketsCallback()                                                                                                                                 
    @     0x55b2ec5b0960  ares__send_query                                                                                                                                                                          
    @     0x55b2ec5b1b55  read_udp_packets                                                                                                                                                                          
    @     0x55b2ec5b1c82  processfds                                                                                                                                                                                
    @     0x55b2ec592e6b  util::fb2::DnsResolve()                                                                                                                                                                   
    @     0x55b2ec55012f  util::http::Client::Reconnect()                                                                                                                                                           
    @     0x55b2ec550ad0  util::http::TlsClient::Connect()                                                                                                                                                          
    @     0x55b2ec1e54a9  dfly::(anonymous namespace)::VersionMonitor::RunTask()                                                                                                                                    
    @     0x55b2ec1e63e7  _ZN5boost7context6detail11fiber_entryINS1_12fiber_recordINS0_5fiberENS0_21basic_fixedsize_stackINS0_12stack_traitsEEEZN4util3fb26detail15WorkerFiberImplIZN4dfly12_GLOBAL__N_114VersionMon
itor3RunEPNS8_12ProactorPoolEEUlvE_JEEC4IS7_EESt17basic_string_viewIcSt11char_traitsIcEERKNS0_12preallocatedEOT_OSH_EUlOS4_E_EEEEvNS1_10transfer_tE                                                                 
    @     0x7fd2300ee24f  make_fcontext                                                                                                                                                                             
*** SIGABRT received at time=1688021637 on cpu 0 ***                                                                                                                                                                
PC: @     0x7fd22f672a7c  (unknown)  pthread_kill                               
```

chakaz:
Yeah, that's a good idea.
So, to reproduce (I did it on a vm just to not mess up my computer):
* Edit `/etc/resolv.conf` to use `127.0.0.1` as nameserver (the default, at least in my case, was `127.0.0.53`)
* Install node
* Download/clone https://github.com/sh1mmer/dnsserver.js
* Edit `dnsserver.js` to:
    * Listen on port 53 (dns default)
    * Add whatever hostname you'd like to resolve (in our case `version.dragonflydb.io`). This server crashes whenever it is queried a nonexisting hostname :facepalm: 
    * Edit line 159 from `response.header.tc = 0;` to `response.header.tc = 1;` (this causes the client to think that the response is truncated (tc), thus to switch to TCP, retry, use another server, whatever.. it's an error, that's what's important I think)
* Run `sudo node dnsserver.js` to serve DNS requests
* Run Dragonfly

Note that in order to test my changes I had to patch also the [code that checks whether to do a version check](https://github.com/dragonflydb/dragonfly/blob/main/src/server/dfly_main.cc#L197-L209) to always perform the check.

PR #90, opened 2023-06-29
feat(tls-socket): allow tls socket to cache the first bytes received

kostasrim:
This is used such that connections can be downgraded from tls to non-tls


kostasrim:
Do we use any specific formater for helio?

romange:
We have clang format file in the repo, imho.

PR #91, opened 2023-07-02
Fix build issue on Fedora

chakaz:
Fixes https://github.com/dragonflydb/dragonfly/issues/1504

Tested on a Fedora VM, and also locally with Ubuntu to make sure nothing breaks


royjacobson (approved):
Nice, thanks!

PR #96, opened 2023-07-13
make helio compile on macOS

yuelimv:
This patch uses `libcares` installed by homebrew on macOS instead. The previous libcares built from the original source seems have extra dependencies on `-lresolv` when helio is built on a mac, which seems to unnecessary. 

CMakeLists.txt is modified to turn off `libunwind`. An alternative way is to pass a flag to `blaze.sh` or `cmake` inside `blaze.sh`, for instance:

`
./blaze.sh -ninja -release -DWITH_UNWIND=OFF
`

Furthermore, we need to avoid linking against `librt` on macOS. This has just been done in another PR

https://github.com/romange/helio/pull/97

Tested on macOS Big Sur 11.7.7

```
> g++ -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 12.0.5 (clang-1205.0.22.11)
Target: x86_64-apple-darwin20.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
> gcc -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 12.0.5 (clang-1205.0.22.11)
Target: x86_64-apple-darwin20.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```


chakaz:
Hi @yuelimv, thanks for your contribution!
I'm not sure that I follow - why not download and compile cares as a static library? It doesn't compile? Maybe we can patch its makefile somehow, like we do for other archs?

Re/ libunwind - I never really used macos, but I read online that it's possible to install, perhaps we could better detect that somehow instead of always disabling it?

yuelimv:
hi @chakaz 

> Hi @yuelimv, thanks for your contribution! I'm not sure that I follow - why not download and compile cares as a static library? It doesn't compile? Maybe we can patch its makefile somehow, like we do for other archs?
> 
Without this patch, the current implementation download a copy of libcares, build it. But for some reason, the one built on mac always ask for libresolv as a dependency, which I couldn't find the way to install. This will need further investigation to see how the library can be configured not to require such dependency. 

> Re/ libunwind - I never really used macos, but I read online that it's possible to install, perhaps we could better detect that somehow instead of always disabling it?

Yes that is possible... Apple has a proprietary version that ships together with macOS... for instance, I can find it under /opt/local/lib, and it might also be possible to install a copy through mac ports. At least manually editing build.ninja to use the correct library path will make it work. But I felt doing so might require some more changes to cmake files comparing to this change. 

Could you explain the benefit of keeping libunwind on macOS? I actually don't quite understand its full purpose. But according to @romange libunwind current introduces bugs that slows down exception handling... it was turned off in release mode for dragonfly

From what I googled, vanilla libunwind seems does not have macOS support according to discussions on their mailing list. As a matter of fact, I couldn't make it compile on macos.

romange:
Hi @chakaz - I've been discussing about this change with @yuelimv offline. I do not consider libunwind as a required feature. We use it only for improving the quality of CPU profiling. Regarding libcares - I think it's fine to use the distribution library: libcares is mature enough and for now we want to get something simple that works, and using a system lib seems to fit the profle :)

cmake/third_party.cmake
```diff
@@ -332,12 +332,26 @@ add_third_party(
...
-  CMAKE_PASS_FLAGS "-DCARES_SHARED:BOOL=OFF -DCARES_STATIC:BOOL=ON -DCARES_STATIC_PIC:BOOL=ON \
+if(APPLE)
+  message("On macOS, please install c-ares via homebrew: ")
+  message("brew install c-ares")
+  message("The current cmake assumes c-ares 1.19.1. For cmake to find
+  the package, create a file named caresConfig.cmake under:")
+  message("/usr/local/Cellar/c-ares/1.19.1/")
+  
+  add_library(TRDP::cares SHARED IMPORTED)
+  find_package(cares REQUIRED PATHS /usr/local/Cellar/c-ares/1.19.1/)
```
romange, line 343: that looks fragile. Can you somehow provide a path without the specified version?

yuelimv: yes, just made the change, I just realized brew will automatically create cmake config file for libcares under `/usr/local/lib/cmake` ...

romange (approved):
LGTM!
Thank you, Yue!

PR #98, opened 2023-07-13
feat(TlsSocket): add ownership semantics to TlsSocket

kostasrim:
TlsSocket is constructed by a pointer to the next socket, effectively acting as a decorator/proxy object of the underline socket. This however is problematic, because once a connection upgrades to tls, you now need to keep alive two different objects with whose lifetime semantics are bound. Therefore, we make TlsSocket own the underline socket, to avoid decoupling the lifetime of the two sockets.

### DIFF
--- a/util/fiber_socket_base.h
+++ b/util/fiber_socket_base.h
@@ -92,7 +92,7 @@ class FiberSocketBase : public io::Sink, public io::AsyncSink, public io::Source
   //! in process of completing.
   virtual uint32_t CancelPoll(uint32_t id) = 0;
 
-  virtual  bool IsUDS() const = 0;
+  virtual bool IsUDS() const = 0;
   virtual bool IsDirect() const = 0;
 
   using native_handle_type = int;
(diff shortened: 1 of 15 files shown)


kostasrim:
Do not merge this unless my other PR passes all tests. I will comment here :)

util/tls/tls_engine_test.cc
```diff
@@ -333,11 +333,6 @@ TEST_F(SslStreamTest, Write) {
   }
 }
 
-
-TEST_F(SslStreamTest, Socket) {
```
romange, line 337: why did you delete it?

kostasrim: because it defaults construct a `TlsSocket`, this test does nothing :D Also, TlsSocket, is no longer default constructable

util/fiber_socket_base.h
```diff
@@ -93,29 +127,29 @@ class FiberSocketBase : public io::Sink, public io::AsyncSink, public io::Source
...
+  ABSL_MUST_USE_RESULT error_code Listen(unsigned backlog) override;
 
   // Listens on all interfaces. If port is 0 then a random available port is chosen
   // by the OS.
-  ABSL_MUST_USE_RESULT error_code Listen(uint16_t port, unsigned backlog);
+  ABSL_MUST_USE_RESULT error_code Listen(uint16_t port, unsigned backlog) override;
 
   // Listen on UDS socket. Must be created with Create(AF_UNIX) first.
-  ABSL_MUST_USE_RESULT error_code ListenUDS(const char* path, mode_t permissions, unsigned backlog);
+  ABSL_MUST_USE_RESULT error_code ListenUDS(const char* path, mode_t permissions, unsigned backlog)override;
```
royjacobson, line 152: formatting?

kostasrim: That's really weird, 99.9% it's a clang-format bug, because I did run it before I pushed

kostasrim: and speaking of that I will add pre-commit here.

romange (approved):

PR #99, opened 2023-07-15
Add macos build

romange:


base/pmr/arena.h
```diff
@@ -7,15 +7,25 @@
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
+
+#ifdef __clang__
```
royjacobson, line 11: We should `#include <version>` and use
```suggestion
#ifndef __cpp_lib_memory_resource
```
instead, they added it in libc++16 and we can use clang+libstdc++ combination as well.

royjacobson: but there are a couple other places where we do this as well so maybe another PR

yuelimv:
I can compiled this branch (both release and dbg modes) with gcc-13 installed via homebrew... 

```
> g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/local/Cellar/gcc/13.1.0/bin/../libexec/gcc/x86_64-apple-darwin20/13/lto-wrapper
Target: x86_64-apple-darwin20
Configured with: ../configure --prefix=/usr/local/opt/gcc --libdir=/usr/local/opt/gcc/lib/gcc/current --disable-nls --enable-checking=release --with-gcc-major-version-only --enable-languages=c,c++,objc,obj-c++,fortran --program-suffix=-13 --with-gmp=/usr/local/opt/gmp --with-mpfr=/usr/local/opt/mpfr --with-mpc=/usr/local/opt/libmpc --with-isl=/usr/local/opt/isl --with-zstd=/usr/local/opt/zstd --with-pkgversion='Homebrew GCC 13.1.0' --with-bugurl=https://github.com/Homebrew/homebrew-core/issues --with-system-zlib --build=x86_64-apple-darwin20 --with-sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 13.1.0 (Homebrew GCC 13.1.0)
```

Tested on macOS Big Sur 11.7.8 (Intel CPU)

PR #106, opened 2023-07-19
Improve TLS handling.

### LINKED ISSUE #105: Improve TLS performance
We currently perform many calls to SSL_write in `util::tls::TlsSocket::WriteSome`. This has a negative impact on performance and should be improved.

royjacobson:
Close #105.

I hope my changes to the write loop are correct - please take a more deep look at them.

### DIFF
--- a/util/tls/tls_socket.cc
+++ b/util/tls/tls_socket.cc
@@ -203,8 +203,9 @@ io::Result<size_t> TlsSocket::Recv(const io::MutableBytes& mb, int flags) {
 }
 
 io::Result<size_t> TlsSocket::WriteSome(const iovec* ptr, uint32_t len) {
-  // Chosen to be sufficiently smaller than the usual MTU (1500) and a multiple of 64.
-  constexpr size_t kBufferSize = 1408;
+  // Chosen to be sufficiently smaller than the usual MTU (1500) and a multiple of 16.
+  // IP - max 24 bytes. TCP - max 60 bytes. TLS - max 21 bytes.
+  constexpr size_t kBufferSize = 1392;
   io::Result<size_t> ec;
   size_t total_sent = 0;
 
--- a/util/tls/tls_socket.cc
+++ b/util/tls/tls_socket.cc
@@ -203,25 +203,44 @@ io::Result<size_t> TlsSocket::Recv(const io::MutableBytes& mb, int flags) {
 }
 
 io::Result<size_t> TlsSocket::WriteSome(const iovec* ptr, uint32_t len) {
-  DCHECK(engine_);
-  DCHECK_GT(len, 0u);
+  // Chosen to be sufficiently smaller than the usual MTU (1500) and a multiple of 64.
+  constexpr size_t kBufferSize = 1408;
+  io::Result<size_t> ec;
+  size_t total_sent = 0;
 
-  Engine::Buffer dest;
+  while (len) {
+    if (ptr->iov_len > kBufferSize || len == 1) {
+      ec = SendBuffer(Engine::Buffer{reinterpret_cast<uint8_t*>(ptr->iov_base), ptr->iov_len});
+      ptr++;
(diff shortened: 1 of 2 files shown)


util/tls/tls_socket.cc
```diff
@@ -206,14 +206,21 @@ io::Result<size_t> TlsSocket::WriteSome(const iovec* ptr, uint32_t len) {
...
 
-  Engine::Buffer dest{reinterpret_cast<uint8_t*>(ptr->iov_base), ptr->iov_len};
+  Engine::Buffer dest;
   size_t send_total = 0;
 
+  if (len == 1) {
+    dest = {reinterpret_cast<uint8_t*>(ptr->iov_base), ptr->iov_len};
+  } else {
+    write_buffer_.Clear();
+    for (size_t i = 0; i < len; i++) {
```
romange, line 216: what if we got 1MB chunks? 
I think this logic makes sense only when total length of the iovec is small. otherwise you copy your data twice - it's too costly.

romange: looking at the code further, I actually think better solution would be is to avoid calling `MaybeSendOutput` in some cases and buffer already encrypted data inside tls engine.  So right now we call `MaybeSendOutput` after each call to `engine_->Write(dest);`  (i.e. prematurely) but instead we could call when tls enginer buffers are at significant capacity (or full).

romange: for that you need to reorder the logic that handles op_result and calls MaybeSendOutput because you must know if there is more data you potentially need to write.

royjacobson: Fair point about the total length of the iovec, but we should still buffer calls to `Write` so we don't create too many TLS records

romange: Are you saying each call to tls engine creates a separate record?

royjacobson: Pretty sure, yeah (the overhead is around 20-30 bytes IIUC but I haven't tested it)

romange: I found this related thread: https://mta.openssl.org/pipermail/openssl-users/2020-February/011753.html
in that case, a reasonable behaviour would be to continue copying until you reach a size larger than `max tls frame size * alpha` and then flush and continue. A quick google search gave 16K as an answer so I think 4kb or 8kb is a good threshold.  In addition, I do not think iobuf is needed here - you can know in advance how much memory you need (4kb or less) and reserve accordingly. a simple `vector<uint8_t>` should suffice I think.

royjacobson: I ended up doing a smaller size - 1408 - because of the usual MTU size of 1500 plus space for metadata. Also wanted to reduce memory footprint of a connection so I just moved the buffer to the stack. LMK if you think it's unreasonable.

util/tls/tls_socket.cc
```diff
@@ -203,18 +203,44 @@ io::Result<size_t> TlsSocket::Recv(const io::MutableBytes& mb, int flags) {
 }
 
 io::Result<size_t> TlsSocket::WriteSome(const iovec* ptr, uint32_t len) {
+  // Chosen to be sufficiently smaller than the usual MTU (1500) and a multiple of 64.
+  constexpr size_t kBufferSize = 1408;
```
romange, line 207: are you sure that the encrypted stream + tls metadata does not go beyong 1500?

romange (approved):
LGTM

util/tls/tls_socket.cc
royjacobson: Good question :) Checked again and it can go over 1500 if the TCP options field is really large. 
IP headers - 24 bytes
TCP headers - max 60 (typically around 20-30)
TLS headers - 21 bytes
I missed the IP headers size in my previous calculation, so I'm adjusting to 1392 now

PR #108, opened 2023-07-24
include MacOS CI and make helio compile and pass tests on macOS

yuelimv:
Currently, three gtests have to be disabled on MacOS:
`malloc_test` (doesn't build on github macos CI instance)
`UDS` from `accept_server_test`
`ProcReader` from `io_test`
`Timeout` from `fiber_socket_test`


romange:
Hi @yuelimv . Seems we solved most of the problems with this PR. Can you please drive it to completion by avoiding doing tests/checks that are not applicable for APPLE? 

Thanks!

yuelimv:
> Hi @yuelimv . Seems we solved most of the problems with this PR. Can you please drive it to completion by avoiding doing tests/checks that are not applicable for APPLE?
> 
> Thanks!

hi @romange , i have rebased and disabled four tests. Build and test passed on my macbook. Those tests seems to be central, but I guess they contain code that are tightly related to Linux... please take a look to see if we need to get one of those tests work on mac. Otherwise, I felt this could be more or less good to go.

io/CMakeLists.txt
```diff
@@ -3,5 +3,8 @@ cxx_link(io base)
 
 add_library(file ALIAS io)
 
-cxx_test(io_test io LABELS CI)
-cxx_test(file_test io LABELS CI)
\ No newline at end of file
+if (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
romange, line 6: I did not mean to disable the whole test suite. I think it's too much.
I meant to skip specific tests inside, for example, like this:
https://github.com/dragonflydb/dragonfly/blob/62a9313a382fa737bc891d7a874fc37122b37a57/src/server/multi_test.cc#L310

same with fiber_socket_test and accept_server_test . In fact they should work.  What exactly does not work there?

yuelimv: agreed. At the end, just three particular tests need to be skipped. We understood the ProcReader test, the other two tests still need investigation.

romange (approved):

PR #110, opened 2023-07-29
fix: fix non-linux set nonblocking

andydunstall:
Running locally found `EpollSocket::Accept` was just blocking forever and never suspending the fiber - looks like `FD_CLOEXEC` is being set with `F_SETFL` instead of `F_SETFD` which is then breaking `O_NONBLOCK`

@romange this is part of the cause of the deadlock you mentioned (if `Accept` never releases the fiber the accept loop never finishes so `AcceptServer::Stop` blocks in the test tear down) - though theres then another issue where `shutdown` isn't waking up the `accept` fiber so it still blocks `Stop` (will look at that separately)

### DIFF
--- a/util/fiber_socket_base.h
+++ b/util/fiber_socket_base.h
@@ -196,6 +196,10 @@ class LinuxSocketBase : public FiberSocketBase {
   int32_t fd_;
 };
 
+void SetNonBlocking(int fd);
+
+void SetCloexec(int fd);
+
 #if 0
 class SocketSource : public io::Source {
  public:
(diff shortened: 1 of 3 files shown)


util/fiber_socket_base.h
```diff
@@ -196,6 +196,10 @@ class LinuxSocketBase : public FiberSocketBase {
   int32_t fd_;
 };
 
+void SetNonBlocking(int fd);
```
romange, line 199: we call it only from epoll_socket (which is both for linux/epoll and macos/freebsd). Lets declare and implement those methods inside epoll_socket.cc in anonymous namespace.

romange (commented):
lgtm -- good catch! a minor comment.

util/fiber_socket_base.h
andydunstall: Its also used by `LinuxSocketBase::Create` - could update `EpollSocket` to override `Create`?

romange: no, it's fine.

romange (approved):

PR #111, opened 2023-07-30
fix: fix accept server test with epoll

andydunstall:
Fixes running the accept server tests with `epoll` (setting `USE_URING` to false). Before the socket was being closed by the wrong thread so hitting the `EpollProactor::Disarm` assertion (https://github.com/romange/helio/blob/0ec06cf8277ebb2e976afdda4f7304fc64010884/util/fibers/epoll_proactor.cc#L372)

Or is epoll meant to support closing from a different thread?

(Note `kqueue` still blocks on tear down due to the `shutdown` not waking the accept fiber issue)

### DIFF
--- a/util/accept_server_test.cc
+++ b/util/accept_server_test.cc
@@ -61,7 +61,7 @@ class AcceptServerTest : public testing::Test {
   void SetUp() override;
 
   void TearDown() override {
-    client_sock_->Close();
+    client_sock_->proactor()->Await([&] { client_sock_->Close(); });
     as_->Stop(true);
     pp_->Stop();
   }
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -118,6 +118,8 @@ EpollSocket::~EpollSocket() {
 auto EpollSocket::Close() -> error_code {
   error_code ec;
   if (fd_ >= 0) {
+    DCHECK_EQ(GetProactor()->thread_id(), pthread_self());
+
     int fd = native_handle();
     DVSOCK(1) << "Closing socket";
     GetProactor()->Disarm(fd, arm_index_);
--- a/util/uring/uring_socket.cc
+++ b/util/uring/uring_socket.cc
@@ -51,6 +51,8 @@ UringSocket::~UringSocket() {
 auto UringSocket::Close() -> error_code {
   error_code ec;
   if (fd_ >= 0) {
+    DCHECK_EQ(GetProactor()->thread_id(), pthread_self());
+
     DVSOCK(1) << "Closing socket";
 
     int fd = native_handle();


royjacobson:
Can we move the assertion into the `Close()` call itself? @romange

romange:
What assertion?

royjacobson:
The ` DCHECK(pthread_self() == thread_id_);` assertion - it's missing in the io_uring proactor code (or just unneeded?)

romange:
@andydunstall  - @royjacobson  got a point. Let's make it easier to debug such bugs in the future. 
Lets add to UringSocket::Close() 

```cpp
Proactor* p = GetProactor();
DCHECK_EQ(p->thread_id(), pthread_self());
```

romange:
@andydunstall  - regarding shutdown not triggerring an event - it's weird. This is what kqueue docs tell us:
https://man.freebsd.org/cgi/man.cgi?kqueue
```
If	the read direction of the socket has shutdown,
			     then the filter also sets EV_EOF in flags,	and
			     returns the socket	error (if any) in fflags....
```
Can you confirm  that `EpollSocket::Wakey` is not trigerred when shutdown is called on a socket? If it's called, then probably the code there is faulty.

andydunstall:
> Let's make it easier to debug such bugs in the future.

Yep sure - added to `Close` in the `if (fd_ >= 0)` (otherwise destructors cause it to fail)

andydunstall:
> Can you confirm that EpollSocket::Wakey is not trigerred when shutdown is called on a socket? 

`Wakey` is called on the connection socket, but not the listen socket. Running a quick test locally calling `shutdown(fd, SHUT_RDWR)` on an accept socket registered with kqueue/epoll, then calling `kqueue/epoll_wait`, `epoll_wait` returns whereas `kqueue` just blocks (taken from https://habr.com/en/articles/600123/#accepting-socket-connections-with-kqueue). Though I haven't found any reference that says a listen socket won't be woken up so could just be doing something wrong

Kqueue:
```
...
struct kevent events[2];
EV_SET(&events[0], obj.sk, EVFILT_READ, EV_ADD | EV_CLEAR, 0, 0, &obj);
EV_SET(&events[1], obj.sk, EVFILT_WRITE, EV_ADD | EV_CLEAR, 0, 0, &obj);
assert(0 == kevent(kq, events, 2, NULL, 0, NULL));

shutdown(obj.sk, SHUT_RDWR);

printf("BLOCK\n");

// wait for incoming events from kqueue
struct timespec *timeout = NULL; // wait indefinitely
int n = kevent(kq, NULL, 0, events, 1, timeout);
printf("DONE\n");
```

epoll:
```
...
// attach socket to KQ
struct epoll_event event;
event.events = EPOLLIN | EPOLLOUT | EPOLLET;
event.data.ptr = &obj;
assert(0 == epoll_ctl(kq, EPOLL_CTL_ADD, obj.sk, &event));

shutdown(obj.sk, SHUT_RDWR);

printf("BLOCK\r\n");

// wait for incoming events from KQ
struct epoll_event events[1];
int timeout_ms = -1; // wait indefinitely
int n = epoll_wait(kq, events, 1, timeout_ms);
printf("DONE\n");
```

romange:
In that case EVFILT_USER could help. We can send it upon shutdown for
freebsd
and register for it when listening.

andydunstall:
> In that case EVFILT_USER could help. We can send it upon shutdown for freebsd and register for it when listening.

Ok sure - would it be easier for shutdown to just call `Wakey(EpollProactor::EPOLL_IN, 0, nullptr);` (rather than notifying `kqueue` which then calls `Wakey`)?

romange (approved):

PR #118, opened 2023-08-06
Fix bug in EpollSocket::RecvMsg

royjacobson:
Fixes [dragonflydb/dragonfly/1591](https://github.com/dragonflydb/dragonfly/issues/1591)

The problem is that a code path in `EpollSocket::RecvMsg` would not reset `read_context_` so fibers would resume at random suspend points outside.

### DIFF
--- a/util/fibers/fibers_test.cc
+++ b/util/fibers/fibers_test.cc
@@ -388,9 +388,7 @@ TEST_F(FiberTest, Notify) {
 TEST_F(FiberTest, CleanExit) {
   ASSERT_EXIT(
       {
-        thread th([] {
-          Fiber([] { exit(42); }).Join();
-        });
+        thread th([] { Fiber([] { exit(42); }).Join(); });
         th.join();
       },
       ::testing::ExitedWithCode(42), "");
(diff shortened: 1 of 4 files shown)


util/fibers/fibers_test.cc
```diff
@@ -616,5 +614,75 @@ TEST_P(ProactorTest, Mutex) {
     th.reset();
 }
 
+TEST_P(ProactorTest, DragonflyBug1591) {
+  unique_ptr<ProactorThread> ths[kNumThreads];
```
romange, line 618: do you use ths, fbs?

util/fibers/fibers_test.cc
```diff
@@ -616,5 +614,75 @@ TEST_P(ProactorTest, Mutex) {
...
+  Fiber fbs[kNumThreads];
+
+  auto sock = std::unique_ptr<FiberSocketBase>(proactor()->CreateSocket());
+  auto sock2 = std::unique_ptr<FiberSocketBase>(proactor()->CreateSocket());
```
romange, line 622: wow, seems it was not easy to reproduce even when you knew exactly what happens.

romange (commented):
lgtm

util/fibers/fibers_test.cc
royjacobson: no, thanks :)

royjacobson: no, thanks :)

romange (approved):

PR #120, opened 2023-08-21
fix: correctly handle cpu cycle reset events

romange:
Addresses https://github.com/dragonflydb/dragonfly/issues/1705

### DIFF
--- a/util/fibers/proactor_base.cc
+++ b/util/fibers/proactor_base.cc
@@ -5,6 +5,7 @@
 #include "util/fibers/proactor_base.h"
 
 #include <absl/base/attributes.h>
+#include <absl/base/internal/cycleclock.h>
 #include <signal.h>
 
 #if __linux__
@@ -53,7 +54,12 @@ void SigAction(int signal, siginfo_t*, void*) {
   }
 }
 
+inline uint64_t GetCPUCycleCount() {
+  return absl::base_internal::CycleClock::Now();
+}
+
 unsigned pause_amplifier = 50;
+uint64_t cycles_per_10us = 1000000;  // correctly defined inside ModuleInit.
 std::once_flag module_init;
 
 }  // namespace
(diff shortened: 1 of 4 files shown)


util/fibers/proactor_base.cc
```diff
@@ -234,6 +240,17 @@ void ProactorBase::RegisterSignal(std::initializer_list<uint16_t> l, std::functi
   }
 }
 
+void ProactorBase::ProcessSleepFibers(detail::Scheduler* scheduler) {
+  // avoid calling steady_clock::now() too much.
+  // Cycles count can reset, for example when CPU is suspended, therefore we also allow
+  // "returning  into past". False positive is possible but it's not a big deal.
+  uint64_t now = GetCPUCycleCount();
+  if (now < last_sleep_cycle_ || now >= last_sleep_cycle_ + cycles_per_10us) {
```
royjacobson, line 248: Could you add a warning about the time jump?

royjacobson (commented):
So, I'm not sure I follow about the exact reason for the high CPU: it seems we would _never_ execute `ProcessSleep`? But why would that create high CPU usage?

I'm wondering if we might have a similar bug for clock adjustments.

util/fibers/proactor_base.cc
romange: I can - but FYI, this does not affect correctness of this function since `ProcessSleep` uses a real clock anyways.

romange:
> So, I'm not sure I follow about the exact reason for the high CPU: it seems we would _never_ execute `ProcessSleep`? But why would that create high CPU usage?
> 
> I'm wondering if we might have a similar bug for clock adjustments.

I fixed the root cause. The bug that caused the elevated CPU usage is here: https://github.com/romange/helio/blob/master/util/fibers/uring_proactor.cc#L631-L638

`scheduler->NextSleepPoint()` is in the past (because it never gets processed), therefore `ts_arg` points to zero interval causing ` wait_for_cqe` to constantly spin.

util/fibers/proactor_base.cc
romange: Please tell me if you still find this warning useful.

royjacobson: Yeah, seems to me like a thing that might be useful one day and it's rare enough that I don't care about log polution :)

royjacobson:
Ah, clever LOG_FIRST usage :)

royjacobson (approved):

PR #122, opened 2023-08-27
ring_buffer: Add tests coverage and clean-up the interface a bit.

royjacobson:
1. Add testing coverage and a bunch of `[D]CHECK`s
2. Slightly improve (IMO) the exposed functions. In particular, the new `GetTail` allows optional overriding which will allow reusing resources without having to construct a new intermediate object.

Part of the road towards partial sync :)

### DIFF
--- a/base/ring_buffer.h
+++ b/base/ring_buffer.h
@@ -3,69 +3,70 @@
 //
 #pragma once
 
+#include <cassert>
 #include <climits>
 #include <memory>
+#include <utility>
 
 namespace base {
 
 // Simple, single-threaded ring buffer.
+// T must be default constructible.
 template <typename T> class RingBuffer {
  public:
   using value_type = T;
 
-  // cap must be power of 2.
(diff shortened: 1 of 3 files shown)


base/ring_buffer.h
```diff
@@ -5,67 +5,69 @@
 
 #include <climits>
 #include <memory>
+#include <utility>
+
+#include "glog/logging.h"
```
romange, line 10: I usually do no add `glog` in include files, lets use assert instead

base/ring_buffer_test.cc
```diff
@@ -0,0 +1,67 @@
+// Copyright 2013, Beeri 15.  All rights reserved.
```
romange, line 1: please use 
```
// Copyright 2023, Roman Gershman.  All rights reserved.
// See LICENSE for licensing terms.
//

```

royjacobson: oops :)

romange (approved):

PR #125, opened 2023-08-31
feat(http): Add authentication support

royjacobson:
We should probably protect the metrics page somehow, this is the quickest way to do it.

This is currently untested because testing is kind of a hassle (no http unit tests at all currently). However I tested it works locally and I'm willing to write a Python test with `requests` once it's merged :)

### DIFF
--- a/util/http/http_handler.cc
+++ b/util/http/http_handler.cc
@@ -372,10 +372,10 @@ bool HttpConnection::CheckRequestAuthorization(const RequestType& req, HttpConte
 
   bool authenticated = false;
   if (const boost::string_view header = req[h2::field::authorization]; !header.empty()) {
-    if (header.substr(0, 6) == "Basic ") {
-      boost::string_view encoded_login = header.substr(6);
+    std::string_view header_stl{header.data(), header.size()};
+    if (absl::StartsWith(header_stl, "Basic ")) {
       std::string decoded;
-      absl::Base64Unescape({encoded_login.data(), encoded_login.size()}, &decoded);
+      absl::Base64Unescape(header_stl.substr(6), &decoded);
       auto split_pos = decoded.find(':');
       if (split_pos != std::string::npos) {
         std::string username = decoded.substr(0, split_pos);
--- a/util/http/http_handler.cc
+++ b/util/http/http_handler.cc
@@ -387,7 +387,7 @@ bool HttpConnection::CheckRequestAuthorization(const RequestType& req, HttpConte
     }
   }
 
-  if (!owner_->password_.empty() && !authenticated) {
+  if (!authenticated) {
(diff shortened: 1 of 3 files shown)


romange:
I am not sure how prometheus will access it.
Moreover, not sure how humans can access it.

royjacobson:
It's pretty standard. The browser prompt looks like this:
image

I think all the prometheus tooling will support it, but cc @ashotland

util/http/http_handler.cc
```diff
@@ -366,9 +367,44 @@ void HttpConnection::HandleRequests() {
...
+    return false;
+  }
+
+  return true;
+}
+
 void HttpConnection::HandleSingleRequest(const RequestType& req, HttpContext* cntx) {
   CHECK(owner_);
 
+  if (!CheckRequestAuthorization(req, cntx))
```
romange, line 405: lets at least do it only if password is not empty

royjacobson: done

util/http/http_handler.h
```diff
@@ -80,6 +80,12 @@ class HttpListenerBase : public ListenerInterface {
     enable_metrics_ = true;
   }
 
+  // Enforce authentication to this HTTP server with the user "user"
+  // and this password.
+  void SetPassword(std::string password) {
```
chakaz, line 85: Correct me if I'm wrong, but this is dead code, right? (i.e. nothing calls `SetPassword()`)
Where do you intend to call this method?

util/http/http_handler.cc
```diff
@@ -366,9 +367,44 @@ void HttpConnection::HandleRequests() {
   LOG_IF(INFO, !FiberSocketBase::IsConnClosed(ec)) << "Http error " << ec.message();
 }
 
+bool HttpConnection::CheckRequestAuthorization(const RequestType& req, HttpContext* cntx) {
+  CHECK(owner_);
+
+  bool authenticated = false;
+  if (const boost::string_view header = req[h2::field::authorization]; !header.empty()) {
+    if (header.substr(0, 6) == "Basic ") {
```
chakaz, line 375: Use `absl::StartsWith()`?

util/http/http_handler.h
royjacobson: From `facade::Connection::HandleRequests`.

chakaz: Ok, please add `override` then :)

royjacobson: But I now added a demonstration to `http_main`

royjacobson: It's not virtual though?

chakaz: Now I understand, thanks!

chakaz (approved):

PR #129, opened 2023-09-05
feat(fibers): Add utility for dumping the stacktraces of all fibers in a thread

royjacobson:
I've been wanting something like this for a while.

### DIFF
--- a/util/fibers/detail/fiber_interface.cc
+++ b/util/fibers/detail/fiber_interface.cc
@@ -268,13 +268,23 @@ ctx::fiber_context FiberInterface::SwitchTo() {
 }
 
 void FiberInterface::PrintStackTrace() {
-  if (FiberActive() == this) {
-    LOG(INFO) << "------------ Fiber " << name_ << " ------------:\n" << GetStacktrace();
-    return;
+  std::string_view state = "sleeping";
+  if (list_hook.is_linked()) {
+    state = "ready";
+  } else if (FiberActive() == this) {
+    state = "active";
   }
 
-  entry_ = std::move(entry_).resume_with([name = name_](ctx::fiber_context&& c) {
-    LOG(INFO) << "------------ Fiber " << name << " ------------:\n" << GetStacktrace();
+  auto print = [this, state]() {
+    LOG(INFO) << "------------ Fiber " << this->name_ << " (" << state << ") ------------\n" << GetStacktrace();
+  };
+
+  if (FiberActive() == this) {
+    return print();
+  }
+
+  entry_ = std::move(entry_).resume_with([print = std::move(print)](ctx::fiber_context&& c) {
+    print();
 
     c = std::move(c).resume();
     return std::move(c);
--- a/util/fibers/detail/fiber_interface.cc
+++ b/util/fibers/detail/fiber_interface.cc
@@ -240,7 +240,7 @@ void FiberInterface::ActivateOther(FiberInterface* other) {
 }
 
 void FiberInterface::DetachThread() {
-  scheduler_->DetachWorker();
+  scheduler_->DetachWorker(this);
   scheduler_ = nullptr;
 }
 
@@ -268,9 +268,13 @@ ctx::fiber_context FiberInterface::SwitchTo() {
 }
 
 void FiberInterface::PrintStackTrace() {
+  if (FiberActive() == this) {
+    LOG(INFO) << "------------ Fiber " << name_ << " ------------:\n" << GetStacktrace();
+    return;
+  }
+
   entry_ = std::move(entry_).resume_with([name = name_](ctx::fiber_context&& c) {
-    std::string stacktrace = GetStacktrace();
-    LOG(INFO) << "Fiber " << name << ":\n" << stacktrace;
+    LOG(INFO) << "------------ Fiber " << name << " ------------:\n" << GetStacktrace();
 
     c = std::move(c).resume();
     return std::move(c);
--- a/util/fibers/detail/fiber_interface.cc
+++ b/util/fibers/detail/fiber_interface.cc
@@ -269,7 +269,7 @@ ctx::fiber_context FiberInterface::SwitchTo() {
 
 void FiberInterface::PrintStackTrace() {
   entry_ = std::move(entry_).resume_with([name = name_](ctx::fiber_context&& c) {
-    std::string stacktrace = GetStacktrace(false);
+    std::string stacktrace = GetStacktrace();
     LOG(INFO) << "Fiber " << name << ":\n" << stacktrace;
 
     c = std::move(c).resume();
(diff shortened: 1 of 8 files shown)


util/fibers/detail/fiber_interface.cc
```diff
@@ -290,6 +291,20 @@ ctx::fiber_context FiberInterface::SwitchTo() {
   });
 }
 
+void FiberInterface::PrintStackTrace() {
+  entry_ = std::move(entry_).resume_with([name = name_](ctx::fiber_context&& c) {
+    std::string stacktrace = GetStacktrace(false);
```
chakaz, line 296: I'd default to true actually, as I imagine that in most cases that we print stack traces we won't care too much about performance

util/fibers/detail/fiber_interface.cc
```diff
@@ -290,6 +291,20 @@ ctx::fiber_context FiberInterface::SwitchTo() {
...
+    LOG(INFO) << "Fiber " << name << ":\n" << stacktrace;
```
chakaz, line 297: I'd add as much information on the fiber as possible, maybe epoch? I realize we don't have IDs, but perhaps that's also worth adding..

util/fibers/detail/fiber_interface.h
```diff
@@ -64,6 +64,9 @@ class FiberInterface {
   FI_SleepHook sleep_hook;
 
   ::boost::context::fiber_context SwitchTo();
+  void PrintStackTrace();
+
+  static void PrintAllFiberStackTraces();
```
chakaz, line 69: Would we be able to call this upon crash, via `absl::InstallFailureSignalHandler()`? I can see cases in which this might turn useful, but perhaps this will cause another signal to be raised, if we're already crashing?

util/fibers/detail/scheduler.cc
```diff
@@ -638,6 +638,31 @@ void Scheduler::RunDeferred() {
 #endif
 }
 
+void Scheduler::PrintAllFiberStackTraces() {
+  ProcessRemoteReady();
```
chakaz, line 642: Why are you calling this here?

util/fibers/stacktrace.cc
```diff
@@ -0,0 +1,62 @@
...
+#include <dlfcn.h>
+#include <execinfo.h>
+
+#include "absl/strings/str_format.h"
+#include "absl/strings/str_join.h"
+#include "base/logging.h"
+
+std::string util::GetStacktrace(bool symbolize) {
+  void* addresses[20];
+  int size = backtrace(addresses, sizeof(addresses) / sizeof(void*));
```
chakaz, line 16: I don't know if `backtrace()` is portable (like on MacOS / FreeBSD), maybe use [`absl::GetStackTraceWithContext()`](https://github.com/abseil/abseil-cpp/blob/master/absl/debugging/stacktrace.h) instead?

chakaz (commented):
That's a great idea, it will surely be useful in some cases!

romange:
It's super cool! Can it be integrated into gdb?

util/fibers/detail/fiber_interface.cc
romange: We do have names

chakaz: I'm not saying we should rename the name (which Roy's already printing :)

util/fibers/detail/fiber_interface.cc
royjacobson: It's reeeeeally slow though

util/fibers/detail/scheduler.cc
royjacobson: `ProcessRemoteReady()` is pushing fibers into the ready queue if they were scheduled from `ScheduleFromRemote`. The remote queue doesn't support iteration and it's emptied into `ready_queue_` anyway.

util/fibers/detail/fiber_interface.cc
royjacobson: abseil is faster, so enabled everywhere.

util/fibers/detail/fiber_interface.h
royjacobson: Maybe; I'm not super sure how well we can count on everything to still be up :/

util/fibers/detail/scheduler.cc
romange: so please add a comment saying that you call it to gather more fibers that are in the process of being activated into ready queue.

util/fibers/stacktrace.h
```diff
@@ -0,0 +1,13 @@
+// Copyright 2023, Roman Gershman.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#pragma once
+
+#include <string>
+
+namespace util {
+
```
romange, line 10: namespace fb2

util/fibers/detail/scheduler.cc
```diff
@@ -553,11 +555,18 @@ void Scheduler::ScheduleFromRemote(FiberInterface* cntx) {
 void Scheduler::Attach(FiberInterface* cntx) {
   cntx->scheduler_ = this;
 
+  fibers_.push_back(*cntx);
```
romange, line 558: Please add a test involving  `ProactorBase::Migrate` call which migrates the calling fiber to a differerrent (proactor) thread

util/fibers/detail/scheduler.cc
```diff
@@ -662,6 +673,46 @@ void Scheduler::RunDeferred() {
...
+    LOG(INFO) << "Sleeping fibers:";
+    print_sleeping_header = false;
+    for (auto& fiber : sleep_queue_) {
+      print_stack_trace(fiber);
+    }
+  }
+
+  // Print any other fibers that are sleeping outside of the scheduler's reach.
+  // Don't print ourselves?
+  for (auto& fiber : fibers_) {
```
romange, line 705: you do you need so many loops if you have a single list covering all the fibers. Is not it simpler to go over only "fibers_" and deduce fibers state based on their hooks?

util/fibers/detail/scheduler.cc
royjacobson: Since we now just loop over the fibers, this doesn't seem to matter too much anymore so I removed it.

util/fibers/detail/scheduler.cc
royjacobson: There's a `ProactorTest.Migrate` test already, do you want to test specifically the combination of migration and the stack traces?

romange:
Yes, I want to see that the list of fibers is updated correctly after the
migration . in fact, please dcheck that all fibers you traverse have the
correct scheduler from the current thread.

romange (approved):

PR #123, opened 2023-08-29
Add coverage to helio tests

romange:

PR #135, opened 2023-09-07
fix(s3): fix url encoding s3 path

andydunstall:
Fixes https://github.com/dragonflydb/controlplane/issues/225, where if the bucket or key includes a character that AWS expects to be URL encoded, the signature will be invalid

AWS defines their own URL path encoding in https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html:
```
URI encode every byte. UriEncode() must enforce the following rules:
- URI encode every byte except the unreserved characters: 'A'-'Z', 'a'-'z', '0'-'9', '-', '.', '_', and '~'.
- The space character is a reserved character and must be encoded as "%20" (and not as "+").
- Each URI encoded byte is formed by a '%' and the two-digit hexadecimal value of the byte.
- Letters in the hexadecimal value must be uppercase, for example "%1A".
- Encode the forward slash character, '/', everywhere except in the object key name. For example, if the object key name is photos/Jan/sample.jpg, the forward slash in the key name is not encoded.

The standard UriEncode functions provided by your development platform may not work because of differences in implementation and related ambiguity in the underlying RFCs. We recommend that you write your own custom UriEncode function to ensure that your encoding will work.
```

This implementation is ported from the Go SDK https://github.com/aws/aws-sdk-go/blob/main/private/protocol/rest/build.go#L261

### DIFF
--- a/util/cloud/aws.cc
+++ b/util/cloud/aws.cc
@@ -5,6 +5,7 @@
 #include "util/cloud/aws.h"
 
 #include <absl/strings/str_cat.h>
+#include <absl/strings/str_format.h>
 #include <absl/strings/str_join.h>
 #include <absl/strings/str_split.h>
 #include <absl/time/clock.h>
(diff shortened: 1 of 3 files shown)


andydunstall:
Tested with:
- `redis-cli save`
- `redis-cli save backup-{timestamp}`
- `redis-cli save &&&`
- `redis-cli save adsf!@#123`
- `redis-cli save foo:bar`

util/cloud/aws.cc
```diff
@@ -362,6 +363,31 @@ bool IsExpiredBody(string_view body) {
...
+
+// Escapes the given path as documented by
+// https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html.
+std::string AwsEscapePath(const std::string_view& path, bool encode_sep) {
+  std::string escaped;
+  for (char c : path) {
+    if (!AwsIsEscaped(c) || (c == '/' && !encode_sep)) {
+      escaped.push_back(c);
+    } else {
+      escaped += absl::StrFormat("%%%02X", c);
```
romange, line 385: nit, this allocates a string for every escaped character.
you can achieve the same with `absl::StrAppend(&escaped, "%%",  absl::Hex(c, absl::kZeroPad2))`

romange (commented):
lgtm

util/cloud/aws.cc
andydunstall: Trying that `absl::Hex` returns lowercase though we need uppercase to match the signature

romange:
Got ya. Then let's use absl::StrAppendFormat()

andydunstall:
> Then let's use absl::StrAppendFormat()

Ok sure updated

romange (approved):

PR #136, opened 2023-09-10
feat(s3): pass ownership of http client to s3 read file

andydunstall:
When opening a read file, pass ownership of the HTTP client to the read file - this matches the behaviour of `OpenWriteFile`

Can refactor more later to make this a bit cleaner, though adding for now to get loading snapshots from S3 working (otherwise the bucket must not be destructed before the file which is messy to implement when the 'S3' file is used as a generic interface)

### DIFF
--- a/util/cloud/s3.cc
+++ b/util/cloud/s3.cc
@@ -249,7 +249,13 @@ io::Result<io::ReadonlyFile*> S3Bucket::OpenReadFile(string_view path,
   } else {
     full_path = absl::StrCat(bucket_, "/", path);
   }
-  return OpenS3ReadFile(region_, full_path, aws_, http_client_.get(), opts);
+
+  unique_ptr http_client = std::move(http_client_);
+  error_code ec = Connect(http_client->connect_timeout_ms());
+  if (ec)
+    return make_unexpected(ec);
+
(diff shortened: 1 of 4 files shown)


util/cloud/s3.cc
```diff
@@ -249,7 +249,13 @@ io::Result<io::ReadonlyFile*> S3Bucket::OpenReadFile(string_view path,
...
     full_path = absl::StrCat(bucket_, "/", path);
   }
-  return OpenS3ReadFile(region_, full_path, aws_, http_client_.get(), opts);
+
+  unique_ptr http_client = std::move(http_client_);
+  error_code ec = Connect(http_client->connect_timeout_ms());
+  if (ec)
+    return make_unexpected(ec);
+
+  return OpenS3ReadFile(region_, full_path, aws_, http_client.release(), opts);
```
royjacobson, line 258: Can we modify it so `OpenS3ReadFile` receives a `unique_ptr` instead of a raw pointer? This might seem innocent, but this is problematic because when you lose exception safety until you construct the new `unique_ptr` in `S3WriteFile::S3WriteFile`: If an exception happens somewhere between this pointer will be leaked.

andydunstall: Yep sure - will update

royjacobson (approved):
lgtm

PR #137, opened 2023-09-11
Prefer hard-coded ssl certificates location first

romange:
Use SSL_CTX_set_default_verify_paths as a fallback. The motivation is that when we statically link openssl, it tries to load certificates from the folder that existed on the build distribution and it fails loading certificates on the target platform.

### DIFF
--- a/util/http/http_client.cc
+++ b/util/http/http_client.cc
@@ -5,6 +5,7 @@
 #include "util/http/http_client.h"
 
 #include <absl/strings/numbers.h>
+#include <openssl/err.h>
 
 #include <boost/asio/connect.hpp>
 #include <boost/beast/core/flat_buffer.hpp>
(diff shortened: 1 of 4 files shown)


util/http/http_client.cc
```diff
@@ -37,6 +38,72 @@ int VerifyCallback(int ok, X509_STORE_CTX* ctx) {
...
+    const char* func;
+    l = ERR_get_error_all(&file, &line, &func, &data, &flags);
+#else
+    l = ERR_get_error_line_data(&file, &line, &data, &flags);
+#endif
+  } while (l);
+}
+
+// returns -1 if failed to load any CA certificates, 0 if loaded successfully
+static int SslProbeSetDefaultCaLocation(SSL_CTX* ctx) {
```
royjacobson, line 59: Should this be exposed? We have a different function that sets up a client context in `protocol_client.cc:CreateSslClientCntx()`

romange: would you like me to expose it via http client interface? or maybe put into tls code?

royjacobson: I think a free function is fine (probably in some outside header?)

royjacobson (approved):

PR #142, opened 2023-09-18
feat(aws): build aws sdk

andydunstall:
Builds the AWS SDK (not yet integrated)

The SDK recommends using `find_package`, though since we're using `add_third_party` I don't think that will work. So instead this just adds the libraries outputted when running `find_package` locally. Not sure if theres a nicer way to do this

(Note I compared the `s3v2_demo` size from https://github.com/romange/helio/pull/141 with `s3_demo` and they were about the same, so I don't think adding the SDK increases the binary size much)

Running locally building `s3_demo` vs `s3v2_demo` - this does increase build time from around 1 minute to 3 minutes (on a fresh build so not cached)... Unfortunately there's lots of dependencies we only use a small part of but AWS still builds, which we can't really avoid without relatively big changes to the SDK


andydunstall:
Managing dependencies is probably the main argument against using the SDK at this point and instead just copying what we need

Build times aren't great given theres lots of dependencies of which 90%+ seem redundant for what we're doing, plus macos and fedora have odd build issues (eg fedora is writing libraries to `third_party/libs/aws/lib64` when it writes to `third_party/libs/aws/lib` on ubuntu)...

romange (approved):

PR #144, opened 2023-09-19
chore: introduce custom printing function

romange:


util/fibers/detail/fiber_interface.cc
```diff
@@ -275,14 +275,23 @@ void FiberInterface::PrintStackTrace() {
...
-    LOG(INFO) << "------------ Fiber " << this->name_ << " (" << state << ") ------------\n" << GetStacktrace();
+  auto print_fn = [state](FiberInterface* fb) {
+    LOG(INFO) << "------------ Fiber " << fb->name_ << " (" << state << ") ------------\n"
+              << GetStacktrace();
   };
 
+  PrintStackTrace(std::move(print_fn));
+}
+
+void FiberInterface::PrintStackTrace(PrintFn fn) {
```
royjacobson, line 286: Let's rename this into 'TraverseThreadFibers` or something :)

royjacobson, line 286: Let's rename this into `ExecuteInFiber` or something :)

romange: ok, by the way, do you understand what lines 295-300 do? if yes, I would like to add comments there, @royjacobson .

util/fibers/detail/fiber_interface.cc
```diff
@@ -267,24 +266,13 @@ ctx::fiber_context FiberInterface::SwitchTo() {
...
-
+void FiberInterface::ExecuteOnFiberStack(PrintFn fn) {
   if (FiberActive() == this) {
-    return print();
+    return fn(this);
   }
 
-  entry_ = std::move(entry_).resume_with([print = std::move(print)](ctx::fiber_context&& c) {
-    print();
+  entry_ = std::move(entry_).resume_with([fn = std::move(fn), this](ctx::fiber_context&& c) {
```
royjacobson, line 274: Is this comprehensible?

```suggestion
  // We're in a random fiber but we want to execute in the context of `this`. We call
  // `this->entry_.resume_with(L)`. Calling this method suspends the current fiber and executes `L`
  // on top of the stack of `this->entry_`.
  // It is similar to running an interrupt handler in the same stack as a suspended program.
  // Inside `L`, we execute `fn` and then resume the original fiber that was suspended which is
  // passed to us as the argument. When the other fiber will be resumed normally, it will execute
  // the final `return` statement and go back to its original suspension site.
  entry_ = std::move(entry_).resume_with([fn = std::move(fn), this](ctx::fiber_context&& c) {
```

romange: site -> state.

royjacobson (approved):

PR #147, opened 2023-09-25
fix: macos thread blocking in `EpollProactor::PeriodicCb`

andydunstall:
The `read` call at https://github.com/romange/helio/blob/4c261c19f415ffde5ec9af09915d4d3704e330fe/util/fibers/epoll_proactor.cc#L465 blocks forever meaning that thread is just blocked. That `read` makes sense with Linux `timerfd_create` though I don't understand it for `kqueue`

### DIFF
--- a/util/fibers/epoll_proactor.cc
+++ b/util/fibers/epoll_proactor.cc
@@ -461,10 +461,13 @@ void EpollProactor::PeriodicCb(PeriodicItem* item) {
   }
 
   item->task();
+
+#ifdef __linux__
   uint64_t res;
   if (read(item->val1, &res, sizeof(res)) == -1) {
     LOG(ERROR) << "Error reading from timer, errno " << errno;
   }
+#endif
 }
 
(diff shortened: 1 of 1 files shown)


util/fibers/epoll_proactor.cc
```diff
@@ -461,10 +461,13 @@ void EpollProactor::PeriodicCb(PeriodicItem* item) {
   }
 
   item->task();
+
```
romange, line 464: Good catch. I actually added a test `TEST_P(ProactorTest, Periodic)` that should have covered this functionality. I wonder why it has not trigerred any issues.

romange (approved):

util/fibers/epoll_proactor.cc
andydunstall: odd - it passes in CI but running locally it blocks (without this fix) - its doing `read(1, ...)` (which is `stdout`) which maybe is undefined behaviour so different on arm64...?

PR #148, opened 2023-09-26
feat(sock): Add a new cancellation callback interface.

royjacobson:
There's a minimal Dragonfly PR ready as well.

### DIFF
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -45,7 +45,7 @@ nonstd::unexpected<error_code> MakeUnexpected(std::errc code) {
 }
 
 #ifdef __linux__
-constexpr int kEventMask = EPOLLIN | EPOLLOUT | EPOLLET;
+constexpr int kEventMask = EPOLLIN | EPOLLOUT | EPOLLET | EPOLLHUP | EPOLLERR | EPOLLRDHUP;
 
 int AcceptSock(int fd) {
   sockaddr_in client_addr;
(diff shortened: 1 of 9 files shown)


util/fibers/epoll_socket.cc
```diff
@@ -45,7 +45,7 @@ nonstd::unexpected<error_code> MakeUnexpected(std::errc code) {
 }
 
 #ifdef __linux__
-constexpr int kEventMask = EPOLLIN | EPOLLOUT | EPOLLET;
+constexpr int kEventMask = EPOLLIN | EPOLLOUT | EPOLLET | EPOLLHUP | EPOLLERR | EPOLLRDHUP;
```
romange, line 48: see https://man7.org/linux/man-pages/man2/epoll_ctl.2.html

HUP/ERR are always reported

util/fibers/uring_socket.h
```diff
@@ -60,7 +63,9 @@ class UringSocket : public LinuxSocketBase {
   uint8_t register_flag() const {
     return fd_ & REGISTER_FD ? IOSQE_FIXED_FILE : 0;
   }
+
+  uint32_t error_cb_event_ = UINT32_MAX;
```
romange, line 67: maybe call it `error_cb_id_` ?  event can be confused with the eventmask.

util/fibers/epoll_socket.cc
```diff
@@ -426,7 +436,7 @@ bool EpollSocket::SuspendMyself(detail::FiberInterface* cntx, std::error_code* e
 void EpollSocket::Wakey(uint32_t ev_mask, int error, EpollProactor* cntr) {
   DVSOCK(2) << "Wakey " << ev_mask;
 #ifdef __linux__
-  constexpr uint32_t kErrMask = EPOLLERR | EPOLLHUP;
+  constexpr uint32_t kErrMask = EPOLLERR | EPOLLHUP | EPOLLRDHUP;
```
romange, line 439: was there a reason for EPOLLRDHUP ? I think we got the event with `EPOLLHUP` as well.

royjacobson: I get a `0x2005` aka `EPOLLIN | EPOLLOUT | EPOLLRDHUP` mask when the socket closes.

romange (approved):

PR #150, opened 2023-09-27
feat(HttpHandler): allow to skip auth on paths via callback

kostasrim:
* This PR allows the user of `HttpListenerBase` to register a `cb` that if it returns true, it will skip the `auth`.
* Added `username_` field to `HttpListenerBase` such that we can verify against a specific username

@romange I decided to slightly diverge from what we originally discussed, namely that the customization point will be the `auth` callback. For our use case, we don't really need that, as the only thing we want to do, is to skip the `auth` part on certain `paths`. Otherwise, we would need to also expose to the user, functions that deal with parsing of the `http` protocol and for our use case this is not really needed.

### DIFF
--- a/util/http/http_handler.cc
+++ b/util/http/http_handler.cc
@@ -27,7 +27,7 @@ namespace {
 
 using FileResponse = ::boost::beast::http::response<::boost::beast::http::file_body>;
 
-inline std::string_view as_absl(::boost::string_view s) {
+inline std::string_view ToStd(::boost::string_view s) {
   return std::string_view(s.data(), s.size());
 }
 
@@ -245,7 +245,7 @@ HttpListenerBase::HttpListenerBase() {
 }
 
 bool HttpListenerBase::HandleRoot(const RequestType& request, HttpContext* cntx) const {
-  std::string_view target = as_absl(request.target());
+  std::string_view target = ToStd(request.target());
   if (target == "/favicon.ico") {
     h2::response<h2::string_body> resp = MakeStringResponse(h2::status::moved_permanently);
     resp.set(h2::field::location, favicon_url_);
@@ -296,7 +296,7 @@ bool HttpListenerBase::RegisterCb(std::string_view path, RequestCb cb) {
 }
 
 // Limit the parsing buffer to 4K.
-HttpConnection::HttpConnection(const HttpListenerBase* base) : owner_(base), req_buffer_(4096)  {
+HttpConnection::HttpConnection(const HttpListenerBase* base) : owner_(base), req_buffer_(4096) {
 }
 
 error_code HttpConnection::ParseFromBuffer(io::Bytes buf) {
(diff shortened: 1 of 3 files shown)


util/http/http_handler.cc
```diff
@@ -402,15 +402,18 @@ bool HttpConnection::CheckRequestAuthorization(const RequestType& req, HttpConte
 void HttpConnection::HandleSingleRequest(const RequestType& req, HttpContext* cntx) {
   CHECK(owner_);
 
-  if (!owner_->password_.empty() && !CheckRequestAuthorization(req, cntx))
-    return;
+  std::string_view target = as_absl(req.target());
+  std::string_view path, query;
+  tie(path, query) = ParseQuery(target);
+
+  if (owner_->maybe_skip_auth_ && !(*owner_->maybe_skip_auth_)(path)) {
```
kostasrim, line 409: Whoops this will skip if the callback is not registered. Will fix it sec

util/http/http_handler.h
```diff
@@ -80,12 +81,22 @@ class HttpListenerBase : public ListenerInterface {
...
+    username_ = std::move(username);
+  }
+
+  // Enforce authentication to this HTTP server with the user username_
+  // and pass token password_
   void SetPassword(std::string password) {
     password_ = std::move(password);
   }
 
+  void SetSkipAuthForPath(std::function<bool(std::string_view path)> func) {
```
chakaz, line 96: First, I didn't know you could specify parameter names in function signatures. Nice!
Second, I'd try to make this a unified API. Maybe 2 callbacks: `ShouldAuth(path)` and `TryAuth(user, pass)`
WDYT?
I guess that could be in another PR, but it'd be easier to pull helio once.

chakaz: @royjacobson suggested to have a single API: `Auth(path, user, pass)`
It's a much better idea imho :)

kostasrim: We don't really need `TryAuth(user, pass)` since I added `SetUsername` on the listener. That means, that both the `password` and `username` can be set by the consumers of this lib (in our case dragonfly). `SkipAuthForPath` allows to skip authorization on certain paths, e.g, `/metrics` (which is what we need) and `SetUsername` will allow us to auth against a pair of username/password set by the users of `HttpListenerBase`. 

Maybe I am missing something in your suggestion so I am all eyes :eyes: 

p.s. I had to google what WDYT is :rofl: It means what do you think right? :tong

kostasrim:
The tests that fails is not related, and I see it's been failing for a while...

util/http/http_handler.h
royjacobson: With a callback you can allow multiple users to authenticate to the same HTTP endpoint. I thought it's what we wanted for the ACL stuff, no?

kostasrim: @royjacobson Good question and I am sorry I was not clear about this earlier on the daily. I had a discussion with @romange yesterday and we agreed that currently, admin should only be used by the `default` user. There is no current plan to integrate ACL with admin port or HTTP atm

chakaz: It's still a better API for future expansion, no?

kostasrim: I mean yes but I can't really foresee if this is something that we need and I would rather wait than introducing dead code (in a sense that is not used/executed anywhere)

romange: I thought `Auth(path, user, pass)` is what we discussed - being able to whitelist a specific path and to support multiple "user/pass" pairs (where one of them is default/pass). If it somehow was not clear, I am sorry. 
so +1 to a `(path, user, pass)` functor

romange:
Also, please rebase - the last commit should have fixed the CI

util/http/http_handler.h
kostasrim: done

kostasrim:
Do not merge this, I will reply here once I am sure this works well on dragonfly

util/http/http_handler.cc
```diff
@@ -380,7 +382,7 @@ bool HttpConnection::CheckRequestAuthorization(const RequestType& req, HttpConte
       if (split_pos != std::string::npos) {
         std::string username = decoded.substr(0, split_pos);
         std::string password = decoded.substr(split_pos + 1, decoded.size());
-        if (username == "user" && password == owner_->password_) {
+        if (owner_->auth_functor_(path, username, password)) {
```
romange, line 385: this logic won't work to provide public access to  `/metrics`, for example.

kostasrim: Why? I already tested this on DF and it works just fine? 

The callback would contain: 

```
[](auto path, auto username, auto password) {
  if(path == "/metrics") return true;
  //rest of auth here if the above branch is not true
}
```

https://github.com/dragonflydb/dragonfly/pull/1950/files#diff-1a804e4021a32c426f25ba8f4ad590c07f385dc90f01fe277fc9b8886ea90409R20

https://github.com/dragonflydb/dragonfly/pull/1950/files#diff-1a804e4021a32c426f25ba8f4ad590c07f385dc90f01fe277fc9b8886ea90409R20

kostasrim: Is there something I am missing?

romange: you won't reach the callback because the header will be missing for regular requests. `authenticated` will stay false.

romange: you should check how come it passed your tests

kostasrim: Nice catch! 
> you should check how come it passed your tests 
It passed the tests because I explicitly passed `username/password`. I added an extra test to cover this case. 

Again nice catch :pray:

chakaz: You could use the same API (i.e. with `path, user, pass`), and call it twice in the worst case. As long as the callback is fast, and it doesn't have any side effects, it might be an easier solution (or a simpler API)
But your call

util/http/http_handler.cc
```diff
@@ -367,50 +367,55 @@ void HttpConnection::HandleRequests() {
...
+  return false;
 }
 
 void HttpConnection::HandleSingleRequest(const RequestType& req, HttpContext* cntx) {
   CHECK(owner_);
 
-  if (!owner_->password_.empty() && !CheckRequestAuthorization(req, cntx))
+  std::string_view target = ToStd(req.target());
+  std::string_view path, query;
+  tie(path, query) = ParseQuery(target);
```
chakaz, line 411: Any reason not to do:
```cpp
auto [path, query] = ParseQuery(target);
```
?

chakaz (approved):

util/http/http_handler.cc
romange: done

romange (approved):

chakaz (approved):

PR #152, opened 2023-10-05
feat(aws): remove crc32 checksum

andydunstall:
Using the CRC32 header works on MinIO, though running again on S3 before merging the dragonfly changes found S3 returning:
```
Checksum Type mismatch occurred, expected checksum Type: null, actual checksum Type: crc32
```

I don't understand why given it has headers:
```
x-amz-sdk-checksum-algorithm: CRC32\r\n
x-amz-checksum-crc32: uDr/9A==\r\n
```
So I don't know why it expects no checksum... I'm fairly certain it worked before, though maybe I was only testing MinIO (another argument for regression tests using S3 directly)

Will remove for now just to unblock merging the Dragonfly side (it was to improve resource usage rather than a functional requirement) - will look at next

### DIFF
--- a/util/aws/s3_write_file.cc
+++ b/util/aws/s3_write_file.cc
@@ -58,7 +58,8 @@ std::error_code S3WriteFile::Close() {
   for (size_t i = 0; i != parts_.size(); i++) {
     Aws::S3::Model::CompletedPart part;
     part.SetPartNumber(i + 1);
-    part.SetETag(parts_[i]);
+    part.SetETag(parts_[i].etag);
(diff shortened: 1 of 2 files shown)


andydunstall:
(I tested on the AWS SDK with no patches and that seems to fail too...)

romange:
weird. please see `aws s3api put-object --help` using awscli. I remember I used its verbosity/debugging printings to dive deep into the API

andydunstall:
Ah oops I misunderstood how checksums work - you have to include the CRC32 in the 'create upload' request - then each CRC32 part in the final completion request

romange (approved):

PR #153, opened 2023-10-07
fix: expand on a check-fail output that crashed in a test

romange:
see https://github.com/dragonflydb/dragonfly/actions/runs/6436749970/job/17480707509?pr=1998#step:10:1424

We peek into tls engine to see if it produced some output to write into a socket, then we write some and then we consume (drain) the output that we written. The flow is not atomic so it may definitely create errors but unfortunately I still do not understand context well enough to fix the code.


chakaz (approved):

chakaz:
Hey Roman, can you merge these changes please? :pray:

PR #154, opened 2023-10-09
feat(aws): s3 snapshot error handling

andydunstall:
A few improvements to AWS error handling and logging:

**Logger**
The AWS logging is very confusing and verbose. Since we log the returned errors already and can add better error handling, the AWS logs don't add much value so limiting to verbose logs only.

**Remove Unused AWS Client**
Removes the unused AWS client. Note keeping the same `awsv2_lib` name just to keep compatibility with Dragonfly.

**Error Handling**
Improves the error handling and logging of the S3 client since some errors are hard to understand currently (invalid credentials, invalid region, bucket not found...)

**Log Credentials**
Overrides the credentials chain to log which provider was used, or if no provider was found. (It still has the same behaviour as the AWS credential chain provider)


examples/s3_demo.cc
```diff
@@ -1,172 +1,192 @@
...
+      Upload(absl::GetFlag(FLAGS_bucket), absl::GetFlag(FLAGS_key),
+             absl::GetFlag(FLAGS_upload_size), absl::GetFlag(FLAGS_chunk_size));
+    } else if (cmd == "download") {
+      Download(absl::GetFlag(FLAGS_bucket), absl::GetFlag(FLAGS_key));
+    } else {
+      LOG(ERROR) << "unknown command: " << cmd;
+    }
 
-  return 0;
+    util::aws::Shutdown();
```
romange, line 190: you should call `pp->Stop()` and then `return 0` at the end

andydunstall: yep sure

romange (approved):

PR #156, opened 2023-10-13
chore(accept_server): optionally print stack trace on signal

kostasrim:
Allow to optionally print stack traces on received signals

### DIFF
--- a/util/accept_server.h
+++ b/util/accept_server.h
@@ -20,7 +20,8 @@ class AcceptServer {
   void operator=(const AcceptServer&) = delete;
 
  public:
-  explicit AcceptServer(ProactorPool* pool, bool break_on_int = true);
+  explicit AcceptServer(ProactorPool* pool, bool break_on_int = true,
+                        bool print_stack_on_sig = false);
   ~AcceptServer();
 
   void Run();
--- a/util/fibers/accept_server.cc
+++ b/util/fibers/accept_server.cc
@@ -16,11 +16,19 @@ namespace util {
 using namespace boost;
 using namespace std;
 
-AcceptServer::AcceptServer(ProactorPool* pool, bool break_on_int) : pool_(pool), ref_bc_(0) {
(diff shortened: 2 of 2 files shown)


romange:
There are may be several non-intrusive options. 
1. Why not send "DEBUG STACKTRACE" before shutting down the process?
2. There is also `ListenerInterface::PreShutdown` hook which may help with what you need.

kostasrim:
Isn't it the case that we want to print the stacktraces only when `instance.stop()` fails in pytests? There we use `SIGKILL` which brutally kills the process, so I don't think any of the two above get a chance to run? That's why I wanted to a signal, such that before we `proc.kill()` I send a `proc.send_signal(signal.SIGINT)` to allow DF print the stacktraces before it gets brutally killed. (SIGKILL can't be handled or blocked)

Is there something I am missing?

romange:
Ok, if you want to print only after the regular shutdown sequence finishes
then yes, signal is needed. But you can register for another signal in
dragonfly via a proactor. Does not have to be sigint.

Another thing, accessing server logs is the prerequisite for this pr,
because I am not sure the system is stable enough at this stage to actually
execute the stacktrace printing logic.

kostasrim:
> But you can register for another signal in dragonfly via a proactor. Does not have to be sigint.

Yep that's true and now I am thinking about this doesn't need to be done in helio.

kostasrim:
> Another thing, accessing server logs is the prerequisite for this pr, because I am not sure the system is stable enough at this stage to actually execute the stacktrace printing logic.

I agree, this might never happen and it's kinda a longshot. But then what's the value of printing stacktraces then? We want to print them when things go bad such that we can use them to figure out what happened (for cases that the python code must brutally kill DF via `proc.kill()` because it doesn't respond). 

Also regular shutdown is impossible, if we reached `proc.kill` then we don't really have much option. The only room is to issue a signal before we kill and hope that it actually gets processed. 

Another thing that would add value here is the core dump...

PR #158, opened 2023-10-19
feat(aws): add web identity provider

andydunstall:
Add web identity credential provider to chain (needed for k8s integration)


romange (approved):

andydunstall:
With HTTPS support this now works with K8S

PR #159, opened 2023-10-19
fix: pass uds socket flag to peers

dranikpg:
Andy if you read this, please be wary of rebasing onto this once merged because it'll break Dragonfly

Needed for https://github.com/dragonflydb/dragonfly/pull/2033

### DIFF
--- a/util/fibers/fiber_socket_base.cc
+++ b/util/fibers/fiber_socket_base.cc
@@ -190,10 +190,9 @@ auto LinuxSocketBase::Shutdown(int how) -> error_code {
 auto LinuxSocketBase::LocalEndpoint() const -> endpoint_type {
   endpoint_type endpoint;
 
-  if (fd_ < 0)
+  if (fd_ < 0 || IsUDS())
     return endpoint;
 
-  CHECK_EQ(0, fd_ & IS_UDS);
   DCHECK_EQ(0, fd_ & REGISTER_FD);
 
   socklen_t addr_len = endpoint.capacity();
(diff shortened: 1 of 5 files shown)


util/fibers/uring_socket.cc
```diff
@@ -77,7 +77,7 @@ auto UringSocket::Accept() -> AcceptResult {
     int res = accept4(real_fd, NULL, NULL, SOCK_NONBLOCK | SOCK_CLOEXEC);
     if (res >= 0) {
       UringSocket* fs = new UringSocket{nullptr};
-      fs->fd_ = res << kFdShift;
+      fs->fd_ = (res << kFdShift) | (fd_ & kInheritedFlags);
```
dranikpg, line 80: Because we didn't actually pass this flag, IsUDS() was false for all uds peers

romange: I think there is fiber_socket_test or something. Can we cover it there?

dranikpg:
Done. It was the most painful test I've written so far 🙂  (😭)

util/fibers/fiber_socket_base.cc
```diff
@@ -193,7 +193,7 @@ auto LinuxSocketBase::LocalEndpoint() const -> endpoint_type {
   if (fd_ < 0)
     return endpoint;
 
-  DCHECK_EQ(0, fd_ & IS_UDS);
+  CHECK_EQ(0, fd_ & IS_UDS);
```
romange, line 196: for sake of avoiding stupid crashes for no reason during codepaths that are not covered by our tests I suggest returning an empty endpoint like with fd_ < 0

romange:
thanks for writing the test :) I did not remember that we do not have any client side code for UDS 🤷🏼

util/fibers/fiber_socket_base.cc
dranikpg: Sure, did it for RemoteEndpoint() as well. Yet I was planning to review all usages of Endpoint() within Dragonfly as we call them for UDS as well

romange (approved):

PR #163, opened 2023-10-24
feat(aws): add https support

andydunstall:
Adds HTTPS support to the AWS S3 client. (Will also add a dragonfly option to disable/enable HTTPS)

This uses the existing TLS client context and setup (since the AWS HTTP client writes directly to the socket, using `TlsClient` itself would be a pain so it just copies what it needs).

Since the HTTPS client is shared by multiple proactors, this still doesn't do connection caching. This has a bigger overhead with HTTPS than HTTP though is still negligible compared to the 8MB chunk uploads and downloads so I don't think thats an issue (users can disable HTTPS for S3 endpoints if preferred).

(Will add connection caching if/when replacing the AWS SDK and add a separate HTTP client per proactor rather than a global shared client)

**TLS**
Not all AWS endpoints support TLS 1.3, though all support TLS 1.2, therefore updates the SSL context min version to 1.2 (down from 1.3)
  - Testing locally the S3 endpoints don't support 1.3
  - See https://aws.amazon.com/blogs/security/faster-aws-cloud-connections-with-tls-1-3/: *"Over 65% of AWS service API endpoints now support TLS version 1.3. We are continuing work to enable TLS version 1.3 on AWS service API endpoints globally."*

### DIFF
--- a/examples/s3_demo.cc
+++ b/examples/s3_demo.cc
@@ -22,6 +22,7 @@ ABSL_FLAG(std::string, key, "", "S3 file key");
 ABSL_FLAG(std::string, endpoint, "", "S3 endpoint");
 ABSL_FLAG(size_t, upload_size, 100 << 20, "Upload file size");
 ABSL_FLAG(size_t, chunk_size, 1024, "File chunk size");
+ABSL_FLAG(bool, https, false, "Whether to use HTTPS");
 ABSL_FLAG(bool, epoll, false, "Whether to use epoll instead of io_uring");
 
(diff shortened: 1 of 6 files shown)


romange:
You probably told me this but I do not remember now. Are we creating a connection for each request and then we close it?

andydunstall:
> Are we creating a connection for each request and then we close it?

Yeah - not great but given the HTTP client is accessed by multiple proactors seemed like its easiest just to keep it stateless

Given each S3 operation is an 8MB  upload/download the connection overhead is minimal

romange (approved):

PR #165, opened 2023-11-04
chore: track delays when context switching between fibers

romange:
This metric should help us measuring responsiveness of fiber switching.

### DIFF
--- a/util/fibers/detail/scheduler.cc
+++ b/util/fibers/detail/scheduler.cc
@@ -3,6 +3,7 @@
 //
 #include "util/fibers/detail/scheduler.h"
 
+#include <absl/time/clock.h>
 #include <condition_variable>
 #include <mutex>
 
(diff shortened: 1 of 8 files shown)


util/fibers/detail/scheduler.cc
```diff
@@ -416,15 +417,17 @@ void DispatcherImpl::DefaultDispatch(Scheduler* sched) {
...
       FiberInterface* fi = sched->PopReady();
       DCHECK(!fi->list_hook.is_linked());
       DCHECK(!fi->sleep_hook.is_linked());
-      sched->AddReady(this);
+      sched->AddReady(now, this);
 
       DVLOG(2) << "Switching to " << fi->name();
       // qsbr_worker_fiber_online();
-      fi->SwitchTo();
+      fi->SwitchTo(now);
```
chakaz, line 430: This causes the CHECK failure in the test, I guess the CHECK should be GE rather than GT?

chakaz (approved):

PR #170, opened 2023-11-12
Add `SizeTrackingChannel`

chakaz:
This type of channel is very similar to `SimpleChannel`, but it also tracks the size of `T`s it owns.

### DIFF
--- a/util/fibers/simple_channel.h
+++ b/util/fibers/simple_channel.h
@@ -130,6 +130,68 @@ template <typename T, typename Q> void SimpleChannel<T, Q>::StartClosing() {
   pop_ec_.notifyAll();
 }
 
+// SimpleQueue-like interface, but also keeps track over the size of Ts it owns.
+// It has a slightly less efficient TryPush() API as it forces construction of Ts even if they are
+// not pushed.
+// T must have a .size() method, which should return the heap-allocated size of T, excluding
+// anything included in sizeof(T). We could generalize this in the future.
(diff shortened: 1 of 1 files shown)


romange (approved):
approved but how it helps you with measuring the memory usage?

chakaz:
queue capacity * size of object is how much memory is used by the queue, am I missing something?

chakaz:
I see what you may mean, I'll need to track each element's size as we save a string there as well :|

romange:
If this object a string or any other object that uses heap then then the important part is missed. We do not care about 'sizeof string * capacity'

util/fibers/simple_channel.h
```diff
@@ -126,6 +130,68 @@ template <typename T, typename Q> void SimpleChannel<T, Q>::StartClosing() {
   pop_ec_.notifyAll();
 }
 
+// SimpleQueue-like interface, but also keeps track over the size of Ts it owns.
+// It has a slightly less efficient TryPush() API as it forces construction of Ts even if they are
+// not pushed.
+// T must have a .size() method, which should return the heap-allocated size of T, excluding
+// anything included in sizeof(T). We could generalize this in the future.
+template <typename T, typename Queue = folly::ProducerConsumerQueue<T>> class SizeTrackingChannel {
```
romange, line 138: I prefer introducing this in the dragonfly codebase.

chakaz: Ok, will do.
Somewhat ironically, the only change I'll need to make is indeed the capacity() method then :)

PR #173, opened 2023-11-17
fix: optimize intrusive types

romange:
Also, provide additional debugging variable inside connection to allow tracking of its thread assignment.

### DIFF
--- a/util/fibers/detail/fiber_interface.h
+++ b/util/fibers/detail/fiber_interface.h
@@ -7,7 +7,7 @@
 #include <atomic>
 #include <boost/context/fiber.hpp>
 #include <boost/intrusive/set.hpp>
-#include <boost/intrusive/slist.hpp>
+#include <boost/intrusive/list.hpp>
 #include <chrono>
 
 #include "base/mpsc_intrusive_queue.h"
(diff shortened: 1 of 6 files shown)


util/listener_interface.h
```diff
@@ -61,10 +61,6 @@ class ListenerInterface {
   // Connection* are valid only during the call to cb.
   void TraverseConnections(TraverseCB cb);
 
-  // traverses all client connection in current thread. cb must adhere to rules from
-  // `TraverseConnections`;
-  void TraverseConnectionsOnThread(TraverseCB cb);
-
```
dranikpg, line 67: What's wrong with it? I need it to be public for Dragonfly 😢

util/fibers/detail/fiber_interface.h
```diff
@@ -24,7 +24,7 @@ enum class Launch {
 namespace detail {
 
 using FI_ListHook =
-    boost::intrusive::slist_member_hook<boost::intrusive::link_mode<boost::intrusive::safe_link>>;
+    boost::intrusive::list_member_hook<boost::intrusive::link_mode<boost::intrusive::safe_link>>;
```
dranikpg, line 27: 👍🏻 I wonder how much it improves performance for heavy load with frequent disconnects 🙂

dranikpg (approved):

romange:
Ah, I did not know that it's by design

romange:
Can you please fix it, actually?

PR #174, opened 2023-11-19
chore: safe delete

dranikpg:


util/fibers/listener_interface.cc
```diff
@@ -31,39 +31,68 @@ auto native_handle(const Connection& conn) {
...
-      LOG(FATAL) << "inconsistent connection thread id " << c->DEBUG_connlist_pthread_id << " vs "
-                 << pthread_self();
+    auto it = ListType::s_iterator_to(*c);
+
+    {
+      util::ProactorBase *left = nullptr, *right = nullptr;
+      if (auto lit = it; it != list.begin())
+        left = (--lit)->DEBUG_proactor;
+      if (auto rit = it; ++rit != list.end())
+        right = rit->DEBUG_proactor;
```
dranikpg, line 63: > I think we should check it regardless before and log error ...

The only concern I have here is that the comparison with begin() and end() might not be safe if we're already wrongly placed. Boost uses an artificial root node pointer for those for simplicity...

romange (approved):

PR #178, opened 2023-11-23
fix: prevent compiler optimizations related to loading stale tls object

romange:
Specifically, during the function stack migration between threads we may encounter faulty compiler optimizations in the release mode. See https://stackoverflow.com/questions/75592038/how-to-disable-clang-expression-elimination-for-thread-local-variable for more details.


dranikpg (commented):
We also need to check surrounding DF code if it has anything cached

dranikpg (approved):

PR #179, opened 2023-11-24
fix: improve time measurement performance in fibers

romange:
Replace fiber switch latency measurement with cycle clock, assuming that modern CPUs have constant tsc cycle and these measurements are not critical for correctness.

### DIFF
--- a/util/fibers/detail/utils.h
+++ b/util/fibers/detail/utils.h
@@ -1,0 +1,48 @@
+// Copyright 2023, Roman Gershman.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#pragma once
+
+#include <cstdint>
+
+namespace util {
+namespace fb2 {
+
+namespace detail {
+
+inline void CpuPause() {
+#if defined(__i386__) || defined(__amd64__)
+  __asm__ __volatile__("pause");
+#elif defined(__aarch64__)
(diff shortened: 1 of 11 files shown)


util/fibers/detail/utils.h
```diff
@@ -0,0 +1,56 @@
...
+    uint64_t low, high;
+    __asm__ volatile("rdtsc" : "=a"(low), "=d"(high));
+    return (high << 32) | low;
+#elif defined(__aarch64__)
+    uint64_t val;
+    __asm__ volatile("mrs %0, cntvct_el0" : "=r"(val));
+    return val;
+#else
+#error "Unsupported architecture"
+#endif
```
dranikpg, line 39: Do you think that good ol' `clock()` is that slow?

dranikpg (approved):

util/fibers/detail/utils.h
romange: this is our top right now. system call `clock` is relatively slow when we perform fiber context switches many millions time per second. [vdso] is the optimized version of reading the clock,
image

romange: with this change,

image

moreover if I look at the graph, I see that "FiberInterface::Preempt()" CPU usage was cut by more than half.

PR #181, opened 2023-11-27
fix: errorneus check-fail due to redundant tls read

romange:
1. The failure in `TLConnList::List` has been reproduced on aarch64 due to `ProactorBase::me()` call that was substituted by the compiler with the wrong instance. The irony that we use this call to check our invariants.
2. I redesigned the access to ProactorBase::GetIndex and made it accessible from any thread.
3. Removed `DEBUG_proactor` data member and use `socket()->proactor()` instead.
4. Removed the fallback for unlinking in a differrent thread and introduced back the check-fail

### DIFF
--- a/util/fibers/listener_interface.cc
+++ b/util/fibers/listener_interface.cc
@@ -41,63 +41,11 @@ thread_local ListenerInterface::ListenerConnMap ListenerInterface::conn_list;
 struct ListenerInterface::TLConnList {
   ListType list;
   fb2::CondVarAny empty_cv;
+  int pool_index = -1;
 
-  void Link(Connection* c) {
-    DCHECK(!c->hook_.is_linked());
-    c->DEBUG_proactor = ProactorBase::me();
-    if (!list.empty()) {
-      CHECK_EQ(c->DEBUG_proactor, list.front().DEBUG_proactor);
-      CHECK_EQ(c->DEBUG_proactor, list.back().DEBUG_proactor);
-    }
-    list.push_back(*c);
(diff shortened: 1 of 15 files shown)


util/fibers/listener_interface.cc
```diff
@@ -110,6 +58,30 @@ struct ListenerInterface::TLConnList {
   }
 };
 
+void ListenerInterface::TLConnList::Link(Connection* c) {
+  DCHECK(!c->hook_.is_linked());
+  list.push_back(*c);
```
dranikpg, line 63: the absence of underscores for members scares me now 😄

dranikpg (approved):

util/fibers/listener_interface.cc
romange: yeah, need to switch to `this->` notation

util/fibers/proactor_base.h
```diff
@@ -100,14 +100,10 @@ class ProactorBase {
...
-  static int32_t GetIndex() {
-    return tl_info_.proactor_index;
-  }
-
-  // Internal, used by ProactorPool
-  static void SetIndex(uint32_t index) {
-    tl_info_.proactor_index = index;
+  // from Proactor thread. Returns -1 if Proactor is not part of the pool.
+  // Can be accessed from any thread.
+  int32_t GetPoolIndex() const {
```
dranikpg, line 105: The signed'ness of this makes it a little inconvenient, I'd return UINT_MAX :(

PR #184, opened 2023-12-06
chore: fibers improvements

romange:
1. Fix debug crash when a machine suspends and tsc is reset.
2. Improve a fiber test
3. Fix warnings.

### DIFF
--- a/util/fibers/detail/fiber_interface.cc
+++ b/util/fibers/detail/fiber_interface.cc
@@ -4,6 +4,7 @@
 #include "util/fibers/detail/fiber_interface.h"
 
 #include <absl/time/clock.h>
+
 #include <mutex>  // for g_scheduler_lock
 
 #include "base/logging.h"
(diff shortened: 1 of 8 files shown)


util/fibers/detail/fiber_interface.cc
```diff
@@ -267,23 +268,27 @@ ctx::fiber_context FiberInterface::SwitchTo() {
...
-  uint64_t delta_cycles = tsc - prev->cpu_tsc_;
-  if (delta_cycles > g_tsc_cycles_per_ms) {
-    fb_initializer.long_runtime_cnt++;
+  uint64_t tsc = CycleClock::Now();
 
-    // improve precision, instead of "delta_cycles / (g_tsc_cycles_per_ms / 1000)"
-    fb_initializer.long_runtime_usec += (delta_cycles * 1000) / g_tsc_cycles_per_ms;
+  // When a kernel suspends we may get a negative delta because TSC is reset.
+  // We ignore such cases (and they are very rare).
+  if (tsc > cpu_tsc_) {
```
chakaz, line 276: I don't know how the reset is done, so it might be a stupid question, but shouldn't we do `cpu_tsc_ = tsc` regardless of the `if` result (i.e. even if it's `<=`)?

chakaz (approved):

util/fibers/detail/fiber_interface.cc
romange: not stupid at all 🤦

chakaz (approved):

PR #185, opened 2023-12-09
chore: reimplement fiber migrate

romange:
Before, we used a temporary fiber to pass along the migrated fiber to another proactor. Now, we simplify the transition by using the dispatcher fiber that also handles I/O loop. We switch to it and execute the delegation to another proactor.

Also, remove the old code and get rid of sccache that is not very reliable.

PR #189, opened 2023-12-11
chore: reimplement fiber migrate (#185)

romange:
* chore: reimplement fiber migrate

Before, we used a temporary fiber to pass along the migrated fiber to another proactor. Now, we simplify the transition by using the dispatcher fiber that also handles I/O loop. We switch to it and execute the delegation to another proactor.

Also, remove the old code and get rid of sccache that is not very reliable.

---------


romange:
Verified this time that the bug was fixed. See https://github.com/dragonflydb/dragonfly/actions/runs/7177404134

util/fibers/detail/fiber_interface.cc
```diff
@@ -274,6 +274,23 @@ ctx::fiber_context FiberInterface::SwitchTo() {
   });
 }
 
+void FiberInterface::SwitchToAndExecute(std::function<void()> fn) {
+  FiberInterface* prev = SwitchSetup();
+
+  // pass pointer to the context that resumes `this`
+  std::move(entry_).resume_with([prev, fn = std::move(fn)](ctx::fiber_context&& c) {
```
romange, line 281: this is where the bug was. the original line was `entry_ = std::move(....`.
`entry_ = ` part was a weird typo.

chakaz: oh wow!

chakaz (approved):

PR #191, opened 2023-12-11
Add support for drect FDs

romange:


util/fibers/uring_proactor.cc
```diff
@@ -418,45 +412,73 @@ void UringProactor::CancelPeriodicInternal(uint32_t val1, uint32_t val2) {
...
-  DCHECK(register_fd_);
+  if (!direct_fd_)
+    return kInvalidDirectFd;
 
-  if (!register_fd_)
-    return source_fd;
-
-  auto next = std::find(register_fds_.begin() + next_free_fd_, register_fds_.end(), -1);
+  // TODO: to create a linked list from free fds.
+  auto next = std::find(register_fds_.begin() + next_free_index_, register_fds_.end(), -1);
```
romange, line 419: a table that translates direct fds to linux fds.
-1 denotes a free slot.

util/fibers/uring_socket.cc
```diff
@@ -47,17 +50,52 @@ UringSocket::~UringSocket() {
   LOG_IF(WARNING, ec) << "Error closing socket " << ec << "/" << ec.message();
 }
 
+error_code UringSocket::Create(unsigned short protocol_family) {
+  error_code ec = LinuxSocketBase::Create(protocol_family);
```
romange, line 54: we first create a regular socket, then we allocate a direct fd

util/fibers/uring_socket.cc
```diff
@@ -47,17 +50,52 @@ UringSocket::~UringSocket() {
...
+  if (ec) {
+    return ec;
+  }
+
+  UringProactor* proactor = GetProactor();
+  CHECK(proactor && is_direct_fd_ == 0);
+  DCHECK(proactor->InMyThread());
+
+  if (proactor->HasDirectFD()) {
```
romange, line 63: currently this is the only function that actually uses directfd just because accept_server calls it to create a listening socket. This is how I limit its impact on the rest of the code.

util/fibers/uring_socket.cc
```diff
@@ -47,17 +50,52 @@ UringSocket::~UringSocket() {
...
+    fd = proactor->UnregisterFd(direct_fd);
+    if (fd < 0) {
+      LOG(WARNING) << "Error unregistering fd " << direct_fd;
+      return ec;
+    }
+  } else {
+    fd = native_handle();
   }
+
+  posix_err_wrap(::close(fd), &ec);
```
romange, line 96: we must close linux fd as well.

util/fibers/uring_proactor.cc
```diff
@@ -418,45 +412,73 @@ void UringProactor::CancelPeriodicInternal(uint32_t val1, uint32_t val2) {
...
   if (next == register_fds_.end()) {
     size_t prev_sz = register_fds_.size();
+    // enlarge direct fds table.
     register_fds_.resize(prev_sz * 2, -1);
```
chakaz, line 423: Consider adding some `CHECK(!register_fds_.empty())` (or `DCHECK()`?)

chakaz: Actually, `> 1` given we do `prev_sz + 1` below
(I realize you initialize it to 512, but still, it's conditional and I think it's worth having)

chakaz (approved):
My level of understanding this PR: ~60% :nerd_face:

util/fibers/uring_proactor.cc
romange: it can be 1 so that `next_free_index_` will be 2. it's still a valid case to call `find(end, end, val)`

PR #192, opened 2023-12-12
feat: add support for IORING_CQE_F_SOCK_NONEMPTY flag during socket recv

romange:
This applies only for recent version of iouring.
In case the receive buffer was empty (i.e. we have not got IORING_CQE_F_SOCK_NONEMPTY flag) we can save the internal call to recv inside kernel by flagging the receive operation with IORING_RECVSEND_POLL_FIRST.

### DIFF
--- a/util/fibers/fiber_socket_test.cc
+++ b/util/fibers/fiber_socket_test.cc
@@ -12,6 +12,7 @@
 
 #ifdef __linux__
 #include "util/fibers/uring_proactor.h"
+#include "util/fibers/uring_socket.h"
 #endif
 #include "util/fibers/epoll_proactor.h"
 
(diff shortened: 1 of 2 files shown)


util/fibers/fiber_socket_test.cc
```diff
@@ -347,5 +348,45 @@ TEST_P(FiberSocketTest, UDS) {
   LOG(INFO) << "Finished";
 }
 
+#ifdef __linux__
+TEST_P(FiberSocketTest, NotEmpty) {
+  bool use_uring = GetParam() == "uring";
+  bool has_poll_first = false;
+
+
```
chakaz, line 356: Remove double empty line?

util/fibers/fiber_socket_test.cc
```diff
@@ -347,5 +348,45 @@ TEST_P(FiberSocketTest, UDS) {
...
+  if (use_uring) {
+    has_poll_first = reinterpret_cast<UringProactor*>(proactor_.get())->HasPollFirst();
```
chakaz, line 358: Any reason not to use `static_cast`?

util/fibers/fiber_socket_test.cc
```diff
@@ -347,5 +348,45 @@ TEST_P(FiberSocketTest, UDS) {
...
+  }
+
+  if (!has_poll_first) {
+    GTEST_SKIP() << "NotEmpty test is supported only on uring with ";
```
chakaz, line 362:
```suggestion
    GTEST_SKIP() << "NotEmpty test is supported only on uring with poll first";
```

chakaz (approved):

PR #193, opened 2023-12-12
chore: general improvements to fibers

romange:
The main change is that we now rely on liburing function io_uring_peek_batch_cqe
 to peek for completions instead of manually iterating with `io_uring_for_each_cqe` call.
 The reason for this is that iouring API evolved, and the logic for peeking completions became
 non-trivial: it requires calling a system call in some rare cases. So
 in terms correctness and ease of maintainance it's better to rely on the library function to do the right job.
 In addition, fibers scheduler code has non-functional changes:

 a. Added a trace variable to ease with debugging: it shows the last transitioned state of the fiber
     in ambigous situations.
 b. ProcessSleep now returns the number of woken fibers.

### DIFF
--- a/util/fibers/detail/fiber_interface.h
+++ b/util/fibers/detail/fiber_interface.h
@@ -64,7 +64,7 @@ class FiberInterface {
 
   virtual ~FiberInterface();
 
-
+  // Switch functions
   ::boost::context::fiber_context SwitchTo();
   void SwitchToAndExecute(std::function<void()> fn);
 
@@ -184,9 +184,17 @@ class FiberInterface {
   ::boost::context::fiber_context Terminate();
 
   std::atomic<uint32_t> use_count_;  // used for intrusive_ptr refcounting.
-  std::atomic<uint16_t> flags_;
 
+  // trace_ variable - used only for debugging purposes.
+  enum TraceState : uint8_t {
+    TRACE_NONE,
+    TRACE_SLEEP_WAKE,
+    TRACE_TERMINATE,
+    TRACE_READY
+  } trace_ = TRACE_NONE;
   Type type_;
+
+  std::atomic<uint16_t> flags_{0};
 
   // FiberInterfaces that join on this fiber to terminate are added here.
   WaitQueue wait_queue_;
@@ -203,7 +211,7 @@ class FiberInterface {
 
   char name_[24];
 
-private:
+ private:
   // Handles all the stats and also updates the involved data structure before actually switching
   // the fiber context. Returns the active fiber before the context switch.
   FiberInterface* SwitchSetup();
--- a/util/fibers/detail/scheduler.cc
+++ b/util/fibers/detail/scheduler.cc
@@ -10,8 +10,8 @@
 #include <mutex>
 
 #include "base/logging.h"
-#include "util/fibers/stacktrace.h"
 #include "util/fibers/detail/utils.h"
+#include "util/fibers/stacktrace.h"
 
 namespace util {
 namespace fb2 {
@@ -186,7 +186,6 @@ Scheduler::~Scheduler() {
     LOG(FATAL) << "Scheduler is destroyed with " << num_worker_fibers_ << " worker fibers";
   }
 
-
   fibers_.erase(fibers_.iterator_to(*dispatch_cntx_));
   fibers_.erase(fibers_.iterator_to(*main_cntx_));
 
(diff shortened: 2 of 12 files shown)


util/fibers/uring_proactor.cc
```diff
@@ -168,58 +167,82 @@ void UringProactor::Init(unsigned pool_index, size_t ring_size, int wq_fd) {
...
-void UringProactor::DispatchCqe(detail::FiberInterface* current, const io_uring_cqe& cqe) {
-  uint32_t user_data = cqe.user_data & 0xFFFFFFFF;
-  if (user_data >= kUserDataCbIndex) {  // our heap range surely starts higher than 1k.
-    if (ABSL_PREDICT_FALSE(cqe.user_data == UINT64_MAX)) {
-      base::sys::KernelVersion kver;
-      base::sys::GetKernelVersion(&kver);
+void UringProactor::ProcessCqeBatch(unsigned count, io_uring_cqe** cqes,
+                                  detail::FiberInterface* current) {
+  for (unsigned i = 0; i < count; ++i) {
+    const io_uring_cqe& cqe = *cqes[i];
```
romange, line 173: the code of DispatchCqe moved into this loop as is

util/fibers/uring_proactor.cc
```diff
@@ -168,58 +167,82 @@ void UringProactor::Init(unsigned pool_index, size_t ring_size, int wq_fd) {
...
-    return;
-
-  if (user_data == kWakeIndex) {
-    // We were woken up. Need to rearm wakeup poller.
-    DCHECK_EQ(cqe.res, 8);
-    DVLOG(2) << "PRO[" << GetPoolIndex() << "] Wakeup " << cqe.res << "/" << cqe.flags;
+void UringProactor::ReapCompletions(unsigned init_count, io_uring_cqe** cqes,
+                                    detail::FiberInterface* current) {
+  unsigned batch_count = init_count;
+  while (batch_count > 0) {
```
romange, line 232: we fetch completions in batches. Here we make sure we reap all of them.

chakaz:
Hey Roman, shall I take a look despite the failing test? Wanna let me know when to look?

romange:
@chakaz you can look now :)

util/fibers/detail/fiber_interface.h
```diff
@@ -184,10 +184,18 @@ class FiberInterface {
...
   std::atomic<uint32_t> use_count_;  // used for intrusive_ptr refcounting.
-  std::atomic<uint16_t> flags_;
 
+  // trace_ variable - used only for debugging purposes.
+  enum TraceState : uint8_t {
+    TRACE_NONE,
+    TRACE_SLEEP_WAKE,
+    TRACE_TERMINATE,
+    TRACE_READY
+  } trace_ = TRACE_NONE;
```
chakaz, line 194: I'm curious: did you intentionally use this syntax, or did you forget the `;` and clang-format merged the lines?

```suggestion
  };
  trace_ = TRACE_NONE;
```

util/fibers/detail/scheduler.cc
```diff
@@ -325,25 +325,34 @@ void Scheduler::ProcessRemoteReady(FiberInterface* active) {
...
 
-    // Remote thread can add the same fiber exactly once to the remote_ready_queue.
-    // This is why each fiber is removed from its waitqueue and added to the remote queue
-    // under the same lock.
-    DCHECK(!fi->list_hook.is_linked());
-
-    // ProcessRemoteReady can be called by a FiberInterface::PullMyselfFromRemoteReadyQueue
-    // i.e. it is already active. In that case we should not add it to the ready queue.
+    // Generally, fi should not be in the ready queue if it's still in the remote queue.
+    // Because being in the remote queue means fi is still registered in the wait_queue of
```
chakaz, line 329:
```suggestion
    // Generally, fi should not be in the ready queue if it's still in the remote queue,
    // because being in the remote queue means fi is still registered in the wait_queue of
```

util/fibers/detail/scheduler.cc
```diff
@@ -325,25 +325,34 @@ void Scheduler::ProcessRemoteReady(FiberInterface* active) {
...
+    // some event. However, in case fi is waiting with timeout, ProcessSleep below can not
+    // remove fi from the wait_queue and from the remote queue. In that case fi will be put
+    // into special transisitional state by adding it to ready_queue even though it's still
```
chakaz, line 332:
```suggestion
    // into special transitional state by adding it to ready_queue even though it's still
```

chakaz (approved):
Yet another LGTM but I have no idea what I just read :rofl:

util/fibers/detail/fiber_interface.h
romange: that was intentional

PR #195, opened 2023-12-13
chore: add Boost.Fiber acknowledgements

### LINKED ISSUE #194: derivative work of boost.fiber
I think helio is derivative work of boost.fiber.
Your implementation resembles a lot of boost.fiber - for instance the use of linked lists or the idea of a dispatching-fiber inside the scheduler or the kind of queues to handle sleeping/blocked/remote fibers.
You have removed all references to the original work of boost.fiber - including the license - and claims that all of this code is only a result of you thoughts - that is really is annoying and a slap int the face of the authors of boost.fiber.

romange:
Fixes #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,30 @@ in c++17 on Linux systems. The focus is mostly on backend development, data proc
 
 I will gradually add explanations for the most crucial blocks in this library.
 
+## Acknowledgments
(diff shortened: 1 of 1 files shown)


romange:
@olk FYI

chakaz (approved):

PR #196, opened 2023-12-15
feat: enable direct-fd sockets

romange:
1. Fixed socket migration code.
2. Enabled direct-fd feature for server side sockets for kernels that support it.
3. Added some stats about cqe completions to uring_proactor.
4. Added `maxclients` flag to echo_server. 

memtier load: `memtier_benchmark -s 172.31.36.182 --command=PING -t 16 -c 18 -n 60000 --hide-histogram --distinct-client-seed`

## dragonfly on helio master
```
==================================================================================================
Type         Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec
--------------------------------------------------------------------------------------------------
Pings      968745.62         0.30147         0.30300         0.43900         0.51900     19866.85
```

## dragonfly on this PR:
```
==================================================================================================
Type         Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec
--------------------------------------------------------------------------------------------------
Pings     1083269.17         0.28287         0.27900         0.42300         0.51100     22215.48
```

### DIFF
--- a/util/fibers/fiber_socket_base.cc
+++ b/util/fibers/fiber_socket_base.cc
@@ -42,8 +42,8 @@ void FiberSocketBase::SetProactor(ProactorBase* p) {
     proactor_ = nullptr;
   }
   proactor_ = p;
-
-  OnSetProactor();
+  if (p)
+    OnSetProactor();
 }
 
 Result<size_t> FiberSocketBase::Recv(const iovec* ptr, size_t len) {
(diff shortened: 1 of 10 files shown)


util/fibers/fiber_socket_base.cc
```diff
@@ -42,8 +42,8 @@ void FiberSocketBase::SetProactor(ProactorBase* p) {
     proactor_ = nullptr;
   }
   proactor_ = p;
-
-  OnSetProactor();
+  if (p)
+    OnSetProactor();
```
chakaz, line 46: if `p` can be `nullptr`, do we need to add an `else { OnResetProactor(); }` here?

chakaz (approved):

util/fibers/fiber_socket_base.cc
romange: we need to call OnResetProactor when we nullify non-null `proactor_`. at this point `proactor_` is guaranteed to be null.

PR #198, opened 2023-12-20
feat: Add preliminary support to provided buffer rings

romange:
These provided buffer rings can be used by kernel for multishot operations to read the data directly into these buffers. No functionality changes yet, a test will be followed up with socket changes.

Also did some code clean ups on a way with echo server and ping_iouring_server (no functionality changes).

### DIFF
--- a/util/fibers/uring_proactor.cc
+++ b/util/fibers/uring_proactor.cc
@@ -61,6 +61,9 @@ constexpr uint16_t kMsgRingSubmitTag = 1;
 constexpr uint16_t kTimeoutSubmitTag = 2;
 constexpr uint16_t kCqeBatchLen = 128;
 
+constexpr size_t kBufRingEntriesCnt = 8192;
+constexpr size_t kBufRingEntrySize = 64;  // TODO: should be configurable.
+
 }  // namespace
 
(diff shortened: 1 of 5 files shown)


util/fibers/uring_proactor.cc
```diff
@@ -325,6 +347,62 @@ void UringProactor::ReturnRegisteredBuffer(uint8_t* addr) {
...
+
+  int err = 0;
+
+  ring_group.ring = io_uring_setup_buf_ring(&ring_, kBufRingEntriesCnt, group_id, 0, &err);
+  if (ring_group.ring == nullptr) {
+    return -err;  // err is negative.
+  }
+
+  unsigned mask = kBufRingEntriesCnt - 1;
+  ring_group.buf = new uint8_t[kBufRingEntriesCnt * kBufRingEntrySize];
```
chakaz, line 369: Does this buffer need to be aligned for any specific type?

chakaz (approved):

util/fibers/uring_proactor.cc
romange: not that I know of. it's a binary blob that is used to read the data from socket to.

PR #201, opened 2024-01-03
Add rvalue-ref qualifier `str()` to `StringSink`

chakaz:


romange (approved):

romange:
seems that the older gcc version does not allow such overloading. try a different name for the method.

chakaz:
> seems that the older gcc version does not allow such overloading. try a different name for the method.

I'm just an idiot :)
PTAL

romange (approved):

PR #205, opened 2024-01-22
feat: support ReadFixed api with registered buffers

romange:


util/fibers/uring_file.cc
```diff
@@ -281,6 +284,19 @@ std::error_code LinuxFileImpl::Close() {
   return ec;
 }
 
+std::error_code LinuxFileImpl::ReadFixed(io::MutableBytes dest, off_t offset, unsigned buf_index) {
+  FiberCall fc(proactor_);
+  fc->PrepReadFixed(fd_, dest.data(), dest.size(), offset, buf_index);
+  FiberCall::IoResult io_res = fc.Get();
+
```
dranikpg, line 291: I assume in a non prototype version the proactor should know it's buffers and at least dcheck the size

dranikpg (approved):

util/fibers/uring_file.cc
romange: I pulled out registered buffers management from proactor, so it's upto its user to manage them.

PR #206, opened 2024-01-30
feat: introduce async interface for the LinuxFile

romange:
Also, kill virtual inheritance for LinuxFile and just make it concrete. Finally a trivial test is added for covering some of the uring_file code.

### DIFF
--- a/io/io.h
+++ b/io/io.h
@@ -21,6 +21,10 @@ inline Bytes Buffer(std::string_view str) {
   return Bytes{reinterpret_cast<const uint8_t*>(str.data()), str.size()};
 }
 
+template<size_t N> inline MutableBytes MutableBuffer(char (&buf)[N]) {
+  return MutableBytes{reinterpret_cast<uint8_t*>(buf), N};
+}
+
 inline std::string_view View(Bytes bytes) {
   return std::string_view{reinterpret_cast<const char*>(bytes.data()), bytes.size()};
 }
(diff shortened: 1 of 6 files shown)


util/fibers/uring_file.cc
```diff
@@ -261,42 +234,6 @@ Result<size_t> WriteFileImpl::WriteSome(const iovec* v, uint32_t len) {
   return res;
 }
 
-LinuxFileImpl::~LinuxFileImpl() {
```
romange, line 264: this code is just moved below and exposed via `LinuxFile` directly

util/fibers/uring_file.cc
```diff
@@ -384,6 +360,23 @@ error_code LinuxFile::Read(const iovec* iov, unsigned iovcnt, off_t offset, unsi
   return io::ApplyExactly(iov, iovcnt, std::move(cb));
 }
 
+void LinuxFile::ReadFixedAsync(io::MutableBytes dest, off_t offset, unsigned buf_index,
```
romange, line 363: these are new functions.

io/io.h
```diff
@@ -21,6 +21,10 @@ inline Bytes Buffer(std::string_view str) {
   return Bytes{reinterpret_cast<const uint8_t*>(str.data()), str.size()};
 }
 
+template<size_t N> inline MutableBytes MutableBuffer(char (&buf)[N]) {
+  return MutableBytes{reinterpret_cast<uint8_t*>(buf), N};
+}
```
dranikpg, line 26: accepting absl span was also an option, but I see this code is used only for tests

dranikpg (approved):

PR #207, opened 2024-01-31
fix: late event completions should not call user callbacks

romange:
Late event completions arrive even after CancelOnErrorCb was called. This PR fixes this by introducing a refcounted auxillary record that handles the state for completions even if socket was destroyed.

### DIFF
--- a/util/fibers/uring_socket.h
+++ b/util/fibers/uring_socket.h
@@ -76,7 +76,7 @@ class UringSocket : public LinuxSocketBase {
 
   struct ErrorCbRefWrapper {
     uint32_t error_cb_id = 0;
-    uint32_t ref_count = 2;
+    uint32_t ref_count = 2;   // one for the socket reference, one for the completion lambda.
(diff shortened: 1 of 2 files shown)


util/fibers/uring_socket.h
```diff
@@ -74,7 +74,28 @@ class UringSocket : public LinuxSocketBase {
     fd_ = (val << kFdShift) | (fd_ & ((1 << kFdShift) - 1));
   }
 
-  uint32_t error_cb_id_ = UINT32_MAX;
+  struct ErrorCbRefWrapper {
+    uint32_t error_cb_id = 0;
+    uint32_t ref_count = 2;
```
chakaz, line 79: Can you add a comment explaining why `2`? "1 for the callback and 1 for the member" or something like that? I hope I get it

chakaz (approved):

util/fibers/uring_socket.h
romange: done

PR #208, opened 2024-02-01
chore: get rid of clang warnings

romange:


util/aws/http_client.cc
```diff
@@ -94,8 +94,9 @@ std::shared_ptr<Aws::Http::HttpResponse> HttpClient::MakeRequest(
     buf_body = dynamic_cast<boost::interprocess::bufferstream*>(request->GetContentBody().get());
     // Only copy if we don't know the type.
     if (!buf_body) {
-      int content_size;
-      absl::SimpleAtoi(request->GetContentLength(), &content_size);
+      unsigned content_size;
+      CHECK(absl::SimpleAtoi(request->GetContentLength(), &content_size));
```
chakaz, line 98: Are you sure you want to crash if the request is invalid?

util/aws/http_client.cc
```diff
@@ -230,7 +231,8 @@ io::Result<std::unique_ptr<FiberSocketBase>> HttpClient::Connect(const std::stri
   std::error_code ec = socket->Connect(ep);
   if (ec) {
     LOG(WARNING) << "aws: http client: failed to connect; host=" << host << "; error=" << ec;
-    socket->Close();
+    auto ec2 = socket->Close();
+    (void)ec2;
```
chakaz, line 235: You _could_ do:
```suggestion
    std::ignore = = socket->Close();
```
I'm not saying should _should_ though :rofl:

chakaz: (same for other places in this file)

chakaz (approved):

util/aws/http_client.cc
romange: I do not know. This function does not return an error code and the previous code was UB.

util/aws/http_client.cc
romange: oh wow!

util/aws/http_client.cc
chakaz: :shrug:

chakaz (approved):
lol, that's a lot of `std::ignore`s :rofl:

PR #209, opened 2024-02-01
chore: only waitqueue-listed fibers can be notified remotely

romange:


dranikpg (approved):
I see it's only read for a check, so it doesn't fix the underlying bug?

romange:
if you have any ideas - would love to hear

PR #210, opened 2024-02-01
enable macos ci

romange:
1. Fix cmake flags so that we could run with sanitizers.
   Note: due to fibers the address sanitizers is practically useless but
   we will fix it in subsequent PRs.
   The problem was that we need to compile abseil lib with the same debug flags like helio
   so I moved those settings to before we include third_party libs.
2. Fix periodic timer cancelation on macos.
   In general with epoll/kevent we can assume (for now) all the completions stop syncrhonously
   Once we cancel/delete the timer so we can just explicitly delete "item" in epoll_proactor.
   For io_uring the cancelation causes callback to run one more time so it will delete its item.
3. The rest is minor built/test fixes in fibers_test related to macos.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
 
   build-macos:
     runs-on: macos-latest
-    if: ${{ false }}  # Disabled due to mysterious segfaults in fibers_test ✨
+    if: ${{ true }}  #  ✨
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -99,7 +99,7 @@ jobs:
 
   build-macos:
     runs-on: macos-latest
-    if: ${{ false }}  # Disabled due to mysterious segfaults in fibers_test ✨
+    if: ${{ true }}  #  ✨
```
chakaz, line 102: we can remove this line now :)

chakaz (approved):

chakaz (approved):

PR #214, opened 2024-02-11
fix: disable stack protector that does not work well with gperftools

romange:
1.  `readelf --debug-dump=info ./echo_server | grep "DW_AT_producer"` shows that helio is compiled  with stack protection flags on my laptop,  even in O2 mode. Seems like some distro tweaking of gcc because I have not seen it before. As a result, when CPU profiling is enabled, dragonfly can crash with a message "stack smashing detected" . The stacktrace shows the failure happens inside `CpuProfiler::prof_handler` which is a signal handler of gperftools. Seems like a bad interaction of signal handling and stack protection code. Therefore, I disabled these stack flags for thirdpary libraries.
2. Unrelated, I added more logs inside fiber code for a check-fail that sometimes happens in dragonfly unit tests like this one: https://github.com/dragonflydb/dragonfly/actions/runs/7862807475/job/21452699281?pr=2570#step:10:6509

### DIFF
--- a/base/mpsc_intrusive_queue.h
+++ b/base/mpsc_intrusive_queue.h
@@ -65,7 +65,7 @@ template <typename T> class MPSCIntrusiveQueue {
     T* head = head_;
     T* next = MPSC_intrusive_load_next(*head);
 
-    return stub() == head && next == nullptr;
+    return reinterpret_cast<const T*>(&storage_) == head && next == nullptr;
   }
 };
 
--- a/cmake/third_party.cmake
+++ b/cmake/third_party.cmake
@@ -15,7 +15,8 @@ option (WITH_UNWIND "Enable libunwind support" ON)
 set(THIRD_PARTY_LIB_DIR "${THIRD_PARTY_DIR}/libs")
 file(MAKE_DIRECTORY ${THIRD_PARTY_LIB_DIR})
 
-set(THIRD_PARTY_CXX_FLAGS "-std=c++14 -O3 -DNDEBUG -fPIC")
+set(THIRD_PARTY_CXX_FLAGS "-std=c++14 -O3 -DNDEBUG -fPIC -fno-stack-protector \
+    -fno-stack-clash-protection")
 
 find_package(Threads REQUIRED)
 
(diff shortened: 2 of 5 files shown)


base/mpsc_intrusive_queue.h
```diff
@@ -65,7 +65,7 @@ template <typename T> class MPSCIntrusiveQueue {
     T* head = head_;
     T* next = MPSC_intrusive_load_next(*head);
 
-    return stub() == head && next == nullptr;
+    return reinterpret_cast<const T*>(&storage_) == head && next == nullptr;
```
romange, line 68: stub is non-const version `reinterpret_cast<T*>(&storage_)`

cmake/third_party.cmake
```diff
@@ -15,7 +15,8 @@ option (WITH_UNWIND "Enable libunwind support" ON)
 set(THIRD_PARTY_LIB_DIR "${THIRD_PARTY_DIR}/libs")
 file(MAKE_DIRECTORY ${THIRD_PARTY_LIB_DIR})
 
-set(THIRD_PARTY_CXX_FLAGS "-std=c++14 -O3 -DNDEBUG -fPIC")
+set(THIRD_PARTY_CXX_FLAGS "-std=c++14 -O3 -DNDEBUG -fPIC -fno-stack-protector \
```
romange, line 18: remove stack protector

util/fibers/detail/fiber_interface.cc
```diff
@@ -298,8 +298,12 @@ void FiberInterface::PullMyselfFromRemoteReadyQueue() {
   // Therefore we process all the items and since this function is called after the fiber
   // was pulled from the wait queue, it is guaranteed that other threads won't add this object
   // back to the remote_ready_queue.
-  scheduler_->ProcessRemoteReady(this);
-  CHECK(!IsScheduledRemotely());
+  bool res = scheduler_->ProcessRemoteReady(this);
+  if (IsScheduledRemotely()) {
+    LOG(FATAL) << "Failed to pull " << name_ << " from remote_ready_queue. res=" << res
```
romange, line 303: LOG(FATAL) instead of check-fail

chakaz (approved):

base/mpsc_intrusive_queue.h
chakaz: ok, now I understand what you mean :)
please ignore then.

PR #215, opened 2024-02-12
fix: mpsc intrusive queue now reports its precise state during Pop

romange:
Before, we could get "no element" response, when in fact the queue is not empty. This can happen when `tail_` is being in the middle of the update but its "next" field has not been updated yet (inside Push).

Before the change, we treated this state as "Empty", while the queue could have multiple entries (probably 1 or 2). After this change: we retry the pop operation in scheduler until we reach the true depletion of the remote queue.

### DIFF
--- a/base/mpsc_intrusive_queue.h
+++ b/base/mpsc_intrusive_queue.h
@@ -6,6 +6,7 @@
 
 #include <atomic>
 #include <cstddef>
+#include <utility>
 
 namespace base {
 
@@ -53,12 +54,26 @@ template <typename T> class MPSCIntrusiveQueue {
     MPSC_intrusive_store_next(item, nullptr);
     T* prev = tail_.exchange(item, std::memory_order_acq_rel);
 
-    // link the previous tail to the new tail. Also a potential blocking point.
+    // link the previous tail to the new tail.
+    // (*) Also a potential blocking point!
+    // For more details see the linked article above!
     MPSC_intrusive_store_next(prev, item);
   }
 
-  // Poops the first item at the head or returns nullptr if the queue is empty.
-  T* Pop() noexcept;
+  // Pops the first item at the head or returns nullptr if the queue is empty.
+  T* Pop() noexcept {
+    while (true) {
+      auto [elem, empty] = PopWeak();
+      if (elem || empty)
+        return elem;
+    }
+    return nullptr;
+  }
+
+  // Returns an item at the head if exists and a status whether the queue is empty.
+  // There can be a state where the queue is in the middle of Push transaction
+  // and Pop can not pull the element yet. In that case, PopWeak returns {null, false}.
+  std::pair<T*, bool> PopWeak() noexcept;
 
   // Can be run only on a consumer thread.
   bool Empty() const noexcept {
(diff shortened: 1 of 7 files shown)


base/mpsc_intrusive_queue.h
```diff
@@ -53,12 +53,26 @@ template <typename T> class MPSCIntrusiveQueue {
...
     T* prev = tail_.exchange(item, std::memory_order_acq_rel);
 
-    // link the previous tail to the new tail. Also a potential blocking point.
+    // link the previous tail to the new tail.
+    // (*) Also a potential blocking point!
+    // For more details see the linked article above!
     MPSC_intrusive_store_next(prev, item);
   }
 
-  // Poops the first item at the head or returns nullptr if the queue is empty.
```
chakaz, line 60: lol :hankey:

base/mpsc_intrusive_queue.h
```diff
@@ -53,12 +53,26 @@ template <typename T> class MPSCIntrusiveQueue {
...
-  T* Pop() noexcept;
+  // Pops the first item at the head or returns nullptr if the queue is empty.
+  T* Pop() noexcept {
+    while (true) {
+      auto [elem, res] = PopWeak();
+      if (elem || res)
+        return elem;
+    }
+    return nullptr;
```
dranikpg, line 69: res is a confusing name for empty

util/fibers/detail/scheduler.cc
```diff
@@ -320,14 +320,18 @@ bool Scheduler::WaitUntil(chrono::steady_clock::time_point tp, FiberInterface* m
...
-    FiberInterface* fi = remote_ready_queue_.Pop();
-    if (!fi)
-      break;
+    auto [fi, qempty] = remote_ready_queue_.PopWeak();
+    if (!fi) {
+      if (qempty)
+        break;
+      DVLOG(1) << "Retrying ";
+      continue;
+    }
```
dranikpg, line 329: Why do you use PopWeak here if you already fixed Pop to have retries? For the DVLOG?

dranikpg (approved):

util/fibers/detail/scheduler.cc
```diff
@@ -319,15 +319,21 @@ bool Scheduler::WaitUntil(chrono::steady_clock::time_point tp, FiberInterface* m
 
 bool Scheduler::ProcessRemoteReady(FiberInterface* active) {
   bool res = false;
+  [[maybe_unused]] unsigned iteration = 0;
   while (true) {
-    FiberInterface* fi = remote_ready_queue_.Pop();
-    if (!fi)
-      break;
+    auto [fi, qempty] = remote_ready_queue_.PopWeak();
```
chakaz, line 324: Any reason not to use `Pop()` here?

base/mpsc_intrusive_queue.h
romange: fixed

util/fibers/detail/scheduler.cc
romange: yes, also wanted to see how many times we iterate per attempt

base/mpsc_intrusive_queue.h
romange: We are at this stage at home

util/fibers/detail/scheduler.cc
romange: to print retries

chakaz (approved):

chakaz (approved):

PR #219, opened 2024-02-23
chore: avoid building tests and examples in uring project

romange:
Also, check socket on shutdown.

### DIFF
--- a/util/fibers/listener_interface.cc
+++ b/util/fibers/listener_interface.cc
@@ -348,7 +348,7 @@ uint32_t ListenerInterface::GetMaxClients() const {
 }
 
 void Connection::Shutdown() {
-  auto ec = socket_->Shutdown(SHUT_RDWR);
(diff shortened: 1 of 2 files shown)


util/fibers/listener_interface.cc
```diff
@@ -348,7 +348,7 @@ uint32_t ListenerInterface::GetMaxClients() const {
 }
 
 void Connection::Shutdown() {
-  auto ec = socket_->Shutdown(SHUT_RDWR);
+  auto ec = CHECK_NOTNULL(socket_)->Shutdown(SHUT_RDWR);
```
chakaz, line 351: Just making sure - this is unrelated to the third_pary.cmake changes you've made above right?

chakaz (approved):

util/fibers/listener_interface.cc
romange: sorry, upodated the PR description. merging

PR #220, opened 2024-03-01
chore: Add support for ubuntu24 image

romange:
1. Added devcontainer with ubuntu24/gcc13
2. Configured with using a custom boost library that uses ucontext implementation with asan enabled.
3. Fixed fibers code to support ucontext (we never supported yet and always used boost fcontext implementation)
4. Made sure that fibers_test pass under ASAN - fixed the leakage bug when accepting a socket.

### DIFF
--- a/util/fibers/detail/scheduler.cc
+++ b/util/fibers/detail/scheduler.cc
@@ -71,6 +71,9 @@ DispatcherImpl::DispatcherImpl(ctx::preallocated const& palloc, ctx::fixedsize_s
   entry_ = ctx::fiber(std::allocator_arg, palloc, salloc,
                       [this](ctx::fiber&& caller) { return Run(std::move(caller)); });
   scheduler_ = sched;
+#if defined(BOOST_USE_UCONTEXT)
+  entry_ = std::move(entry_).resume();
+#endif
 }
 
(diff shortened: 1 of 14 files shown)


util/fibers/fibers_test.cc
```diff
@@ -387,6 +387,7 @@ TEST_F(FiberTest, AsyncEvent) {
   done.Wait();
 }
 
+#if 0
```
romange, line 390: This test does not work with ASAN, disabled it since it's not very interesting anyhow

util/fibers/detail/scheduler.cc
```diff
@@ -80,14 +83,12 @@ DispatcherImpl::~DispatcherImpl() {
...
 
 ctx::fiber DispatcherImpl::Run(ctx::fiber&& c) {
-  if (c) {
-    // We context switched from intrusive_ptr_release and this object is destroyed.
-    return std::move(c);
-  }
-
-  // Normal SwitchTo operation.
+#if defined(BOOST_USE_UCONTEXT)
+    std::move( c).resume();
```
chakaz, line 87:
```suggestion
    std::move(c).resume();
```

chakaz (approved):

PR #222, opened 2024-03-05
feat: allow fiber stack size customization

romange:
Change the default stack size for connection fibers from 132KB to 64KB


chakaz (approved):
Adding an ability to modify the stack size is great, but it shouldn't help in reducing RSS, right? Because we saw that only touched pages are really accounted in RSS, and if the pages are touched it means that they are in use and we can downsize the stack, am I missing something?

romange:
No, you are not missing anything. I hope by agressively reducing the stack size in development, we be able to identify callchains that use too much of stack memory and fix them.

PR #224, opened 2024-03-10
chore: add IORING_OP_MADVISE support to iouring

romange:
Also add a microbenchmark measuring mmap and madvise calls.

a single madvise call takes around 100ns while mmap is x3 more expensive.

### DIFF
--- a/util/fibers/submit_entry.h
+++ b/util/fibers/submit_entry.h
@@ -27,7 +27,7 @@ class SubmitEntry {
 
   void PrepOpenAt(int dfd, const char* path, int flags, mode_t mode) {
     PrepFd(IORING_OP_OPENAT, dfd);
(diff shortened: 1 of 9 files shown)


util/fibers/submit_entry.h
```diff
@@ -194,14 +194,22 @@ class SubmitEntry {
     sqe_->cancel_flags = flags;
   }
 
-  io_uring_sqe* sqe() {
-    return sqe_;
+  void PrepMadvise(void* addr, off_t len, int advice) {
```
chakaz, line 197: Where is this called?

romange: It's not called in helio. It's part of helio's iouring API

chakaz: gotcha

chakaz (approved):

PR #225, opened 2024-03-10
chore: Refactor BlockingCounter

dranikpg:
* Splits BlockingCounter/Impl into BlockingCounter and EmbeddedBlockingCounter, so the latter can be re-used
* Adds DCHECKs
* Adds clear comments to headers

### DIFF
--- a/util/fibers/synchronization.h
+++ b/util/fibers/synchronization.h
@@ -370,6 +370,8 @@ class Done {
 };
 
 // Use `BlockingCounter` unless certain that the counters lifetime is managed properly.
+// Because the decrement of Dec() can be observed before notify is called, the counter can be still
+// in use even after Wait() unblocked.
 class EmbeddedBlockingCounter {
   const uint64_t kCancelFlag = (1ULL << 63);
 
--- a/util/fibers/synchronization.h
+++ b/util/fibers/synchronization.h
@@ -412,7 +412,8 @@ class EmbeddedBlockingCounter {
 
 // A barrier similar to Go's WaitGroup for tracking remote tasks.
 // Internal smart pointer for easier lifetime management. Pass by value.
-struct BlockingCounter {
+class BlockingCounter {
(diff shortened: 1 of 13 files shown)


romange:
Please rebase @dranikpg

util/fibers/synchronization.h
```diff
@@ -368,104 +369,58 @@ class Done {
...
+};
 
-  // Blocks while the blocking counter is pending.
-  // Returns true on success (reaching 0), false when cancelled.
-  bool Wait() {
-    return impl_->Wait();
-  }
+// A barrier similar to Go's WaitGroup for tracking remote tasks.
+// Internal smart pointer for easier lifetime management. Pass by value.
+struct BlockingCounter {
```
romange, line 415: it was a class before. if we forward declare - can cause warnings

util/fibers/synchronization.h
```diff
@@ -368,104 +369,58 @@ class Done {
...
+  BlockingCounter(unsigned start_count);
 
-  // Blocks while the blocking counter is pending our timeout is reached.
-  // Returns true on success (reaching 0), false when cancelled or timed out.
-  bool WaitFor(const std::chrono::steady_clock::duration& duration) {
-    auto tp = std::chrono::steady_clock::now() + duration;
-    return impl_->WaitUntil(tp);
+  EmbeddedBlockingCounter* operator->() {
```
romange, line 418: this will require adjusting the whole dragonfly codebase due to interface change.
I suggest you introduce wrappers to preserve the previous interface

romange (commented):
LGTM

util/fibers/synchronization.h
dranikpg: It's only 3 or 4 usages, I already fixed them. Copying all functions because of this is weird

romange (approved):

PR #226, opened 2024-03-11
chore: add DEBUG_Count() to barrier

dranikpg:
Forgot to add this one 😅 Need it for DCHECKs. DEBUG prefix is used mainly in DF to indicate it's better not used in real code


chakaz (approved):
LG but I'm not sure my approval is enough?

romange (approved):

PR #227, opened 2024-03-11
chore: add retries when pulling from the remote queue

romange:


adiholden (approved):

util/fibers/detail/scheduler.cc
```diff
@@ -323,15 +323,21 @@ bool Scheduler::ProcessRemoteReady(FiberInterface* active) {
...
+              __asm__ __volatile__("pause");
+              ++iteration;
+              continue;
+            }
           }
         }
         break;
       }
-      DVLOG(1) << "Retrying " << iteration++;
+      iteration++;
```
kostasrim, line 339: Out of curiosity, why `iteration` is marked as `maybe_unused` above?

PR #228, opened 2024-03-13
chore: introduce worker fiber stack size tracking

romange:
I used the following script to track RSS usage vs VM usage for fibers:
https://github.com/ARM-software/CSAL/blob/master/coresight-tools/cskern/pagemap.py

### DIFF
--- a/util/fibers/detail/scheduler.cc
+++ b/util/fibers/detail/scheduler.cc
@@ -68,6 +68,7 @@ DispatcherImpl* MakeDispatcher(Scheduler* sched) {
 DispatcherImpl::DispatcherImpl(ctx::preallocated const& palloc, ctx::fixedsize_stack&& salloc,
                                detail::Scheduler* sched) noexcept
     : FiberInterface{DISPATCH, 0, "_dispatch"} {
+  stack_size_ = palloc.sctx.size;
   entry_ = ctx::fiber(std::allocator_arg, palloc, salloc,
                       [this](ctx::fiber&& caller) { return Run(std::move(caller)); });
   scheduler_ = sched;
@@ -269,6 +270,7 @@ void Scheduler::Attach(FiberInterface* cntx) {
   fibers_.push_back(*cntx);
 
   if (cntx->type() == FiberInterface::WORKER) {
+    worker_stack_size_ += cntx->stack_size();
     ++num_worker_fibers_;
   }
 }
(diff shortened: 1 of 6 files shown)


util/fibers/fiber2.h
```diff
@@ -109,6 +109,12 @@ uint64_t FiberLongRunSumUsec() noexcept;
 // It is advised to call this function when a program starts.
 void SetDefaultStackResource(PMR_NS::memory_resource* mr, size_t default_size = 64 * 1024);
 
+// Returns the total size for worker fibers for the current thread.
+size_t WorkerStackSize();
```
chakaz, line 113:
```suggestion
// Returns the total stack size for worker fibers for the current thread.
size_t WorkerFibersStackSize();
```

chakaz: Maybe it's also worth clarifying that maybe only part of this size is actually allocated and present in rss

util/fibers/detail/scheduler.cc
```diff
@@ -269,19 +270,22 @@ void Scheduler::Attach(FiberInterface* cntx) {
...
   if (cntx->type() == FiberInterface::WORKER) {
+    worker_stack_size_ += cntx->stack_size();
     ++num_worker_fibers_;
   }
 }
 
 void Scheduler::DetachWorker(FiberInterface* cntx) {
   fibers_.erase(fibers_.iterator_to(*cntx));
   --num_worker_fibers_;
+  worker_stack_size_ -= cntx->stack_size();
```
chakaz, line 281: I assume this is called in the migration path, right? (i.e. that we don't leak)

chakaz (approved):

util/fibers/detail/scheduler.cc
romange: yes

util/fibers/fiber2.h
```diff
@@ -109,6 +109,14 @@ uint64_t FiberLongRunSumUsec() noexcept;
 // It is advised to call this function when a program starts.
 void SetDefaultStackResource(PMR_NS::memory_resource* mr, size_t default_size = 64 * 1024);
 
+// Returns the total stack size (virtual memory) for worker fibers for the current thread.
+// Please note that RSS memory usage is usually smaller, depending on the actuall stack usage
+// of the fibers.
+size_t WorkerStackSize();
```
chakaz, line 115: I'd rename this to `WorkerFibersStackSize()` (I proposed that in the previous iteration but perhaps you've missed it because I added another proposal :)

chakaz (approved):

PR #229, opened 2024-03-16
Introduce a test that reproduces data race for fiber notify code 

romange:


util/fibers/synchronization.h
```diff
@@ -388,7 +387,11 @@ class EmbeddedBlockingCounter {
   }
 
   // Returns true on success (reaching 0), false when cancelled. Acquire semantics
-  bool Wait();
+  bool Wait() {
+    uint64_t cnt;
+    ec_.await(WaitCondition(&cnt));
+    return (cnt & kCancelFlag) == 0;
+  }
```
dranikpg, line 394: But why 😭

dranikpg (approved):

util/fibers/synchronization.h
romange: What you did not like? I am open to changes 🙂

dranikpg: It's just a nit 😅 I've put all of the functions into the source file so the header is more readable and compact

romange: Added a retry mechanism, for possible remote queue bug.
Also, moved some large functions to the implementation section below in the header file.

romange:
PTAL

dranikpg (approved):

PR #232, opened 2024-03-20
fix: Do not crash if connection is not found in list

chakaz:
This happens in Dragonfly if we kill the server immediately after starting replication.

Fixes https://github.com/dragonflydb/dragonfly/issues/2737


adiholden (approved):

romange:
@chakaz  can you explain how this happens? based on the code flow it should not  happen

chakaz:
I agree and I don't _know_. I think that we have a place where we started yielding where we didn't before. I didn't think it's worth it to delve into it, but let me know if you think otherwise.

romange:
I would prefer adding DCHECK that `it != end` so we would at least get better stack traces

chakaz:
We already have a stack trace, see https://github.com/dragonflydb/dragonfly/issues/2737
the problem is that because it's a callback on another thread we don't know the origin

PR #235, opened 2024-03-24
fix: allow parsing multiple requests from the same http connection

romange:


util/http/http_handler.cc
```diff
@@ -315,7 +315,7 @@ error_code HttpConnection::ParseFromBuffer(io::Bytes buf) {
   cntx.set_user_data(user_data_);
 
   while (!buf.empty()) {
-    ParserType parser{std::move(request)};
+    ParserType parser;
```
chakaz, line 318: Is `request` not used by the `parser`?

romange: parser returns an parsed requests few lines below. without this change, the parser has reused already parsed requests for parsing it again, so it's good I added multiple requests to the tests - they uncovered the bug.

romange: see line 325

chakaz: gotcha, nice!

chakaz (approved):

PR #237, opened 2024-04-02
feat(io): AlignedBufPool

dranikpg:


io/aligned_buf_pool.cc
```diff
@@ -0,0 +1,99 @@
+#include "io/aligned_buf_pool.h"
+
```
romange, line 2: headers

io/aligned_buf_pool.h
```diff
@@ -0,0 +1,64 @@
...
+struct AlignedBufPool {
+  const static size_t kAlign = 4096;
+
+  // Request aligned buffer. Returns buffer and cleanup that recycles the buffer once dropped.
+  static auto Request(size_t length) {
+    AlignedBuf buf = RequestChunk(length);
+    return std::make_pair(buf, absl::MakeCleanup([ptr = buf.buf.data()] { ReturnChunk(ptr); }));
+  }
+
+  // Initialize buffer on current thread and register backing in io_uring
```
romange, line 38: fibers2 lib depends on io lib, not the other way around. it's better to move this file into fibers dir

io/aligned_buf_pool.cc
```diff
@@ -0,0 +1,99 @@
...
+namespace io {
+
+namespace {
+
+thread_local AlignedBufPool tl_buf_pool;
+
+}  // namespace
+
+void AlignedBufPool::Init(size_t backing_size) {
```
romange, line 16: I think it's a bit mixed with both OS stuff and data structure stuff. 
if we have a separate class, that acts as an allocator (resource manager) why not make it independent from the os stuff?

i.e. Init provides a bank of memory registered and allocated somewhere and the class does not really care about iouring.

io/aligned_buf_pool.cc
```diff
@@ -0,0 +1,99 @@
...
+void AlignedBufPool::ReturnChunk(void* ptr) {
+  if (auto offset = tl_buf_pool.OffsetByPtr(ptr)) {
+    tl_buf_pool.ReturnSegment(*offset);
+  } else {
+    ::operator delete[](ptr, std::align_val_t(kAlign));
+  }
+}
+
+std::optional<size_t /* offset */> AlignedBufPool::RequestSegment(size_t length) {
+  // If possible, squeeze right before first taken segment
```
romange, line 64: I understand the code. We will have to see if it's robust enough - i.e. the uilization of the pool is high.

Regarding the style, please handle `borrowed_.empty()` case first - it will clean up the rest of the code.

Also, please expose used/capacity accessors as well as borrowed length.

io/aligned_buf_pool.cc
```diff
@@ -0,0 +1,99 @@
...
+  size_t tail_mark = borrowed_.empty() ? 0 : borrowed_.back().first + borrowed_.back().second;
+  if (buffer_size_ - tail_mark >= length)
+    return borrowed_.emplace_back(tail_mark, length).first;
+
+  return std::nullopt;
+}
+
+void AlignedBufPool::ReturnSegment(size_t offset) {
+  auto it =
+      std::lower_bound(borrowed_.begin(), borrowed_.end(), std::pair<size_t, size_t>{offset, 0});
```
romange, line 78: we use it for I/O, so we can assume the access pattern leans to the left. 
you can use it by rewriting lower_bound (it's few more lines) and check first `k` first items on the left before jumping to the middle in each iteration.

io/aligned_buf_pool.cc
```diff
@@ -0,0 +1,99 @@
...
+  DCHECK(it != borrowed_.end());
+
+  it->second = 0;  // clear length
+
+  while (!borrowed_.empty() && borrowed_.front().second == 0)
```
romange, line 83: i would refine the condition to check if `it == begin()` or `it+1= end` just to avoid polluting CPU caches

io/aligned_buf_pool.cc
```diff
@@ -0,0 +1,99 @@
...
+}
+
+AlignedBuf AlignedBufPool::RequestChunk(size_t length) {
+  std::optional<size_t> offset = tl_buf_pool.RequestSegment((length + kAlign - 1) / kAlign);
+
+  uint8_t* ptr = nullptr;
+  if (offset)
+    ptr = &tl_buf_pool.buffer_[*offset * kAlign];
+  else
+    ptr = new (std::align_val_t(kAlign)) uint8_t[length];
```
romange, line 49: following the suggestion above, I would also remove the fallback code. 
This implementation enforces no-backpressure policy. I think it's not the right place to commit to this.

io/aligned_buf_pool.h
```diff
@@ -0,0 +1,64 @@
...
+  const std::optional<unsigned> registered_idx;  // optional buf_index for prep_read/write_fixed
+};
+
+// Hands out cunks of a large registered buffer, falls back to allocating non-registered
+// chunk if no free space is available.
+struct AlignedBufPool {
+  const static size_t kAlign = 4096;
+
+  // Request aligned buffer. Returns buffer and cleanup that recycles the buffer once dropped.
+  static auto Request(size_t length) {
```
romange, line 33: the length is always aligned and you always return a buffer of the specified length. 
if you follow the rest of my suggestions, you can just return `unique_ptr<void*, Deleter>` and I think it's gonna be a more strightforward interface.

io/aligned_buf_pool.cc
romange: and it you make it completely abstract, it can reside in helio/base actually.

dranikpg:
In favour of #239

PR #238, opened 2024-04-02
fix: fibers/socket bugs

romange:
1. Prevent from FiberCall to wake up before its waker actually has been called.
2. Fix DnsResolve code to prevent notifying the wrong FiberCall on epoll event.

### DIFF
--- a/util/fibers/uring_proactor.cc
+++ b/util/fibers/uring_proactor.cc
@@ -40,6 +40,8 @@ using detail::SafeErrorMessage;
 
 namespace fb2 {
 
+static_assert(sizeof(FiberCall) == 48);
+
 using detail::FiberInterface;
 
 namespace {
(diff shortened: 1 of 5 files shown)


util/fibers/uring_proactor.cc
```diff
@@ -842,18 +846,35 @@ FiberCall::FiberCall(UringProactor* proactor, uint32_t timeout_msec) : me_(detai
...
+  // was issued by this fiber call. However, it is possible and in fact in some cases we do that
+  // that we issue asynchronously an io_uring request A (could be any call supported by io_uring)
+  // and then issue another request B via FiberCall. Once completion A is being fullfilled, it may
+  // wake this suspended fiber via its completion callback. In that case this fiber will wake up
+  // even though the waker of this call was not run yet.
+  // To avoid this, we suspend until we make sure our own waker runs.
+  do {
+    me_->Suspend();
+  } while (!was_run_);
+
```
dranikpg, line 873: I still suggest to fix dns-resolve suspend and keep a DCHECK here

dranikpg (approved):

dranikpg (approved):

PR #239, opened 2024-04-03
feat(base+util=io): Uring buffer pooling

dranikpg:
Restructed replacement of #237


util/fibers/uring_buf_pool.cc
```diff
@@ -0,0 +1,78 @@
...
+}
+
+UringBuf UringBufPool::Request(size_t length) {
+  length = (length + kAlign - 1) / kAlign * kAlign;
+
+  if (auto offset = segments_.Request(length / kAlign))
+    return {io::MutableBytes{backing_ + *offset * kAlign, length}, buf_idx_};
+
+  return {{new (std::align_val_t(kAlign)) uint8_t[length], length}, std::nullopt};
+}
```
dranikpg, line 68: I remember you asked not to allocate as a fallback? Do you still hold that view? 🙂 I mean, there are only two options - allocating and backpressure

romange: The fallback policy belongs to higher levels. This layer should not prevent whatever policy we choose in dragonfly.

util/fibers/uring_buf_pool.cc
```diff
@@ -0,0 +1,78 @@
...
+
+#include "base/logging.h"
+#include "util/fibers/proactor_base.h"
+#include "util/fibers/uring_proactor.h"
+
+namespace util::fb2 {
+
+namespace {
+
+thread_local UringBufPool tl_buf_pool;
```
romange, line 19: Why not make it a uring proactor member?

dranikpg: Actually, good idea

base/segment_pool.cc
```diff
@@ -0,0 +1,44 @@
...
+  // Otherwise, try appending after last occupied segment
+  size_t tail_mark = taken_.back().first + taken_.back().second;
+  if (size_ - tail_mark >= length)
+    return taken_.emplace_back(tail_mark, length).first;
+
+  return std::nullopt;
+}
+
+void SegmentPool::Return(unsigned offset) {
+  auto it = std::lower_bound(taken_.begin(), taken_.end(), std::make_pair(offset, 0u));
```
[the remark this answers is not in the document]
dranikpg: Added a front() check, but I don't think it matters at all

util/fibers/uring_proactor.cc
```diff
@@ -310,12 +314,28 @@ SubmitEntry UringProactor::GetSubmitEntry(CbType cb, uint16_t submit_tag) {
   return SubmitEntry{res};
 }
 
-int UringProactor::RegisterBuffers(const struct iovec* iovecs, unsigned nr_vecs) {
-  return io_uring_register_buffers(&ring_, iovecs, nr_vecs);
+int UringProactor::RegisterBuffers(size_t size) {
+  size = (size + 4096 - 1) / 4096 * 4096;
+  buf_pool_.backing = new (std::align_val_t(4096)) uint8_t[size];
```
romange, line 319: i think it's better to use `mmap`  instead of going through our allocator layer.
mi_malloc, for example, allocates `size + alignment` to guarantee alignment, and frankly calling `new` here is redundant given we want 4kb multiple and alignment.

util/fibers/uring_proactor.cc
```diff
@@ -310,12 +314,28 @@ SubmitEntry UringProactor::GetSubmitEntry(CbType cb, uint16_t submit_tag) {
...
+  iovec vec{buf_pool_.backing, size};
+  return io_uring_register_buffers(&ring_, &vec, 1);
 }
 
-int UringProactor::UnregisterBuffers() {
-  return io_uring_unregister_buffers(&ring_);
+std::optional<UringBuf> UringProactor::RequestBuffer(size_t size) {
+  DCHECK(buf_pool_.backing);
+  size = (size + 4096 - 1) / 4096 * 4096;
+  if (auto offset = buf_pool_.segments.Request(size / 4096)) {
```
romange, line 329: please add a comment saying that we use segments to track allocations in 4KB units.
Also, consider introducing a constexpr for 4kb size.

romange (commented):
looks good.

util/fibers/uring_proactor.cc
```diff
@@ -310,12 +311,37 @@ SubmitEntry UringProactor::GetSubmitEntry(CbType cb, uint16_t submit_tag) {
   return SubmitEntry{res};
 }
 
-int UringProactor::RegisterBuffers(const struct iovec* iovecs, unsigned nr_vecs) {
-  return io_uring_register_buffers(&ring_, iovecs, nr_vecs);
+int UringProactor::RegisterBuffers(size_t size) {
+  size = (size + UringBuf::kAlign - 1) / UringBuf::kAlign * UringBuf::kAlign;
+
+  // Use mmap to create the backing. For cleanup we rely just on process termination.
```
romange, line 317: why rely on process termination? it may create problems with unit tests, where we create/destroy lots of proactors.
at least clean it up in the d'tor

romange (approved):

PR #240, opened 2024-04-05
chore: add more debug checks to WaitQueue

romange:


dranikpg (approved):
I really think it would have segfaulted anyways if any of the waiter/cntx variables was null

PR #243, opened 2024-04-08
chore: fix http parsing in case a full request was not sent

romange:


util/http/http_handler.cc
```diff
@@ -357,13 +362,18 @@ void HttpConnection::HandleRequests() {
...
 
+    // h2::read sets parser.eager=true
+    // This function reads from the socket until parser reaches done state or error is encountered.
     h2::read(asa, req_buffer_, parser, ec);
     if (ec) {
       break;
     }
 
+    if (!parser.is_done()) {
+      LOG(DFATAL) << "Incomplete http parsing";
```
chakaz, line 374: I'd expect to see a log level not more than warning, as this might mean that the client is misbehaving or that the connection was dropped?

romange: Look at the comment above. h2::read should not allow this

chakaz: Oh, so the above `if (ec) { ... }` handles a bad request? gotcha.

chakaz (approved):

PR #244, opened 2024-04-08
fix: fix tls socket timeout

BorysTheDev:


util/fibers/epoll_socket.h
```diff
@@ -10,7 +10,7 @@
 namespace util {
 namespace fb2 {
 
-class EpollSocket : public LinuxSocketBase {
+class EpollSocket final : public LinuxSocketBase {
```
romange, line 13: i prefer overriding at method level - it sometimes prevents unit-testing.

util/tls/tls_socket.h
```diff
@@ -46,6 +46,14 @@ class TlsSocket : public FiberSocketBase {
     return next_sock_->IsOpen();
   }
 
+  void set_timeout(uint32_t msec) final override {
```
romange, line 49: have you checked it's working?

BorysTheDev: No, I don't know how to check. If you know how to do it, I will check

romange: there is `examples/https_client_cli.cc` that you can run . add a timeout setting and see if it works.
that's a minimum requirement. You can also introduce a unit test with tls client/server  but that's a bigger chunk of work.

BorysTheDev: I've checked that it works

examples/https_client_cli.cc
```diff
@@ -85,10 +85,11 @@ using OpResult = io::Result<RequestResults, std::string>;
...
 OpResult ConnectAndRead(ProactorBase* proactor, std::string_view host, std::string_view service,
                         const char* resource, SSL_CTX* ssl_ctx) {
-  TlsClient http_client{proactor};
-
-  http_client.set_connect_timeout_ms(2000);
+  
   auto list_res = proactor->Await([&] {
+    TlsClient http_client{proactor};
+
+    http_client.set_connect_timeout_ms(2000);
```
romange, line 92: please expose it with a flag, you can leave the default of 2000.

romange (approved):

PR #246, opened 2024-04-08
chore: Fix the crash in tls shutdown

romange:
The root cause was due to a socket being closed and destroyed inside ListenerInterface::RunSingleConnection
and in parallel the listener was performing the shutdown sequence inside ListenerInterface::RunAcceptLoop().

In fact we had a comment saying it's working because socket.Shutdown did not preempt.
In one of my recent PRs I implemented the logic where TlsSocket informs the peer about the shutdown
which made the call preemptible. As a result the whole management code around this became unsafe.

1. I introduced an intrusive refcnt so that a connection would not be destroyed inside RunSingleConnection
   when RunAcceptLoop still uses it.
2. Changed the iteration loop in the latter tu support preemption.
3. Fixed the Shutdown code to support the case where the tcp socket is closed in the middle of shutdown.
4. Found unrelated migration bug that where we failed to Unlink a connection because the listener was destroyed.
   Now we try to unlink only if the connection indeed is linked inside a connection list.

### DIFF
--- a/util/fibers/listener_interface.cc
+++ b/util/fibers/listener_interface.cc
@@ -170,16 +170,18 @@ void ListenerInterface::RunAcceptLoop() {
   pool_->AwaitFiberOnAll([&](auto* pb) {
     auto it = conn_list.find(this);
     DCHECK(it != conn_list.end());
-    auto* clist = it->second;
+    TLConnList* clist = it->second;
 
-    cur_conn_cnt.fetch_add(clist->list.size(), memory_order_relaxed);
+    // Important to pop from the end because Shutdown may suspend and the list may change
+    while (!clist->list.empty()) {
+      Connection& conn = clist->list.back();
+      clist->list.pop_back();
 
-    // This is correctly atomic because Shutdown() does not block the fiber or yields.
-    // However once we implement Shutdown via io_uring, this code becomes unsafe.
-    for (auto& conn : clist->list) {
+      boost::intrusive_ptr guard(&conn);
       conn.Shutdown();
       DVSOCK(1, conn) << "Shutdown";
     }
+    cur_conn_cnt.fetch_add(clist->list.size(), memory_order_relaxed);
   });
 
   VLOG(1) << "Listener - " << ep.port() << " waiting for " << cur_conn_cnt
@@ -208,7 +210,7 @@ ListenerInterface::~ListenerInterface() {
 void ListenerInterface::RunSingleConnection(Connection* conn) {
   VSOCK(2, *conn) << "Running connection ";
 
-  unique_ptr<Connection> guard(conn);
+  boost::intrusive_ptr<Connection> guard(conn);
 
   ListenerConnMap* conn_map = GetSafeTlsConnMap();
   TLConnList* clist = conn_map->find(this)->second;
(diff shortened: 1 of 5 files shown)


romange:
This fix will warrant another patch

util/fibers/listener_interface.cc
```diff
@@ -170,16 +170,18 @@ void ListenerInterface::RunAcceptLoop() {
...
+      clist->list.pop_back();
 
-    // This is correctly atomic because Shutdown() does not block the fiber or yields.
-    // However once we implement Shutdown via io_uring, this code becomes unsafe.
-    for (auto& conn : clist->list) {
+      boost::intrusive_ptr guard(&conn);
       conn.Shutdown();
       DVSOCK(1, conn) << "Shutdown";
     }
+    cur_conn_cnt.fetch_add(clist->list.size(), memory_order_relaxed);
```
chakaz, line 184: Isn't `clist->list` empty at this point?

util/tls/tls_socket.cc
```diff
@@ -66,6 +67,8 @@ TlsSocket::TlsSocket(FiberSocketBase* next) : TlsSocket(std::unique_ptr<FiberSoc
 }
 
 TlsSocket::~TlsSocket() {
+  // sanity check that all pending ops are done.
+  DCHECK_EQ(state_ & 7, 0);
```
chakaz, line 71: lol why `7`?
Maybe `WRITE_IN_PROGRESS | READ_IN_PROGRESS | SHUTDOWN_IN_PROGRESS`?

util/fibers/listener_interface.cc
romange: no necessarily. Shutdown does not close and removes connections. it just shuts them down and they still need to go through their closing sequence before they are removed from the list

chakaz: But you call `pop_back()` in line 178?

romange: oh, it's a bug and I am not sure how to fix it actually.

romange: ok, provided an awkward fix.

romange: it's an intrusive list, meaning that if an element is not linked then the list was reduced by one element.
Also, `Shutdown` by itself does not cause the element to be removed from the list. Shutdown triggers the finalization flow for the connection, that at the end will remove the connection from the list (at the end of RunSingleConnection function).
But if Shutdown preempts we may or may now have the connection to be unlinked depending how quickly RunSingleConnection finishes its run.

romange: the list will never have new elements so any element that is unlinked reduces the list by one, so no endless loop here

chakaz: Where does it remove from the list? I see that in the original code we used to call `pop_back()` here, but in the latest iteration we don't so how does that conform with:

> it's an intrusive list, meaning that if an element is not linked then the list was reduced by one element.

?

Please excuse my picking, I'm just not sure what I'm missing :)

romange: It's inside RunSingleConnection.
here: https://github.com/romange/helio/pull/246/files#diff-01d997cdbc4d2835ea398fa879f935f6dee87da00e4fdf7ae7264bdd5a98f7f8R255-R257

romange: please note that the goal of the loop is not to empty the list but trigger closing flows, so an element may or may not be removed.

chakaz (approved):
:nerd_face:

chakaz:
No need, it makes sense, I approved this PR :)

PR #247, opened 2024-04-10
feat(dns_resolve): Properly support ipv6

dranikpg:


util/fibers/dns_resolve.cc
```diff
@@ -151,8 +146,9 @@ void ProcessChannel(ares_channel channel, AresChannelState* state, DnsResolveCal
 
 }  // namespace
 
-error_code DnsResolve(string host, uint32_t wait_ms, char dest_ip[], ProactorBase* proactor) {
+error_code DnsResolve(string host, uint32_t wait_ms, absl::Span<char> dest_ip, ProactorBase* proactor) {
```
romange, line 149: I prefer keeping the type as `char dest[]` and update the comments around this function

util/fibers/dns_resolve.cc
```diff
@@ -119,17 +119,12 @@ void DnsResolveCallback(void* ares_arg, int status, int timeouts, hostent* hoste
     return set_error();
   }
 
-  if (hostent->h_addrtype != AF_INET) {
```
romange, line 122: I suggest keeping the whitelist of AF_INET and AF_INET6 (there are other families that we do not support)

util/fibers/dns_resolve.cc
dranikpg: Changed. But why? Span is implicitly constructible from char arrays...

dranikpg:
I don't see any tests for this area, so it's ready

util/fibers/dns_resolve.cc
romange: I consider it an awkward type for public interfaces.

util/fibers/dns_resolve.cc
romange: ping

romange:
@dranikpg have you checked manually that it's working?

dranikpg:
> @dranikpg have you checked manually that it's working?

With a replica on localhost

util/fibers/dns_resolve.cc
dranikpg: I already fixed it, the changediff is just different, reversed conditions

romange (approved):

romange:
@dranikpg you need to sign, imho

PR #250, opened 2024-04-14
chore: Use only IPV4 for localhost dns

dranikpg:


romange:
Please see https://github.com/c-ares/c-ares/issues/70 
The original ares_gethostbyname did not support answering the same query with results containing multiple families.

Therefore, the right way to fix it is to switch to https://c-ares.org/ares_getaddrinfo.html
and then duplicate the logic of 
https://github.com/romange/helio/blob/82b3c9d39d4b0ea99325363e22c052454ef9f104/util/fibers/accept_server.cc#L95-L145

where we iterate over multiple answers and "prefer" the one we want.

dranikpg:
Really bad we don't have proper dns tests, I did testing with DF. However fixing the current http_test would take quite some time

util/http/http_test.cc
```diff
@@ -86,5 +86,15 @@ TEST_F(HttpTest, JsonParse) {
   EXPECT_STREQ("val3", doc["key3"].GetString());
 }
 
+TEST_P(ProactorTest, Dns) {
```
romange, line 89: Test_f HttpTest

util/http/CMakeLists.txt
```diff
@@ -16,8 +16,8 @@ cxx_link(http_client_lib fibers2 http_beast_prebuilt http_utils tls_lib)
 cxx_link(http_main fibers2 html_lib http_server_lib TRDP::mimalloc)
 
 
-#add_library(https_client_lib https_client.cc https_client_pool.cc ssl_stream.cc)
-#cxx_link(https_client_lib proactor_lib absl_variant http_beast_prebuilt)
+# add_library(https_client_lib https_client.cc https_client_pool.cc ssl_stream.cc)
```
romange, line 19: ?

dranikpg:
Oh, sorry, wrong files

romange (approved):

romange:
why did you remove the test? 
also, you can test it with localhost

dranikpg:
> why did you remove the test?

The whole file is outdated, it's not even compiled. I tried to uncomment the cmake stuff and it throws too much errors to be easily fixable

PR #253, opened 2024-04-18
chore: Simpler Future<> without locking

dranikpg:
Introduce simple future to replace boost code

replacing aligned_storage with optional, intrusive_ptr with shared_ptr and a mutex with event-count makes the code ten times smaller 🙂

### DIFF
--- a/util/fibers/future.h
+++ b/util/fibers/future.h
@@ -4,339 +4,35 @@
 
 #pragma once
 
-// This code is a modified version of Boost.Fibers future code.
-//
-#include <boost/intrusive_ptr.hpp>
 #include <memory>
+#include <optional>
 
 #include "util/fibers/synchronization.h"
 
 namespace util {
 namespace fb2 {
 
-enum class future_status { ready = 1, timeout, deferred };
-template <typename R> class Future;
-
-namespace detail {
-
-class shared_state_base {
- private:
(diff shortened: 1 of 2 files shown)


dranikpg:
Actually this is unsafe, need to fix this 😅

util/fibers/future.h
```diff
@@ -4,339 +4,39 @@
...
-    base_type::future_->set_value(value);
-  }
-
-  void swap(Promise& other) noexcept {
-    base_type::swap(other);
-  }
-
-  using base_type::get_future;
+  struct Block {
+    T value{};  // replace with aligned_storage or optional if T is not default-constructible
```
romange, line 35: consider doing this.

util/fibers/future.h
```diff
@@ -4,339 +4,39 @@
...
+    std::atomic_bool has_value = false;
+    util::fb2::EventCount waker{};
+  };
+  std::shared_ptr<Block> block;
```
romange, line 39: if you extend to  `std::variant<std::shared_ptr<Block>, T>` you could provide support for immediate future built-in.

romange (approved):

util/fibers/future.h
dranikpg: I already have StringValue in the main repo

PR #254, opened 2024-04-20
Avoid calling DispatchBrief from the dispatch fiber

romange:


util/fibers/proactor_base.cc
```diff
@@ -213,9 +213,16 @@ void ProactorBase::Migrate(ProactorBase* dest) {
   me->scheduler()->SuspendAndExecuteOnDispatcher([me, dest] {
     me->DetachScheduler();
 
-    dest->DispatchBrief([me] {
-      me->AttachScheduler();
-    });
+    auto cb = [me] {
```
romange, line 216: This can explain the crash in the migrate:
```
scheduler.cc:426] Check failed: dispatch_cntx_->list_hook.is_linked()
```
if a dispatcher was preempted here or some other place, the preemption would succeed in release mode due to DCHECK, and 
the scheduler would switch to another fiber with dispatch_cntx_ not being ready.
if that fiber would decide to migrate we would get this check-fail.

util/fibers/proactor_base.cc
```diff
@@ -213,9 +213,16 @@ void ProactorBase::Migrate(ProactorBase* dest) {
...
+        me->AttachScheduler();
+    };
+
+    // We can not use DispatchBrief because it may block dispatch fiber, which is forbidden.
+    // While this state is theoretically possible but it's very improbable, so we should not reach
+    // usleep in the normal state.
+    while (!dest->EmplaceTaskQueue(std::move(cb))) {
+      usleep(0);
+    };
```
dranikpg, line 225: 1. I wouldn't call move for cb, it's a no-op and formally moving multiple from the same object isn't a good practice
2. Can it deadlock? If two full proactors try to migrate two fibers onto each other. Seems like no because we always migrate via `SuspendAndExecuteOnDispatcher` that doesn't clog the queue?

dranikpg (dismissed):

util/fibers/proactor_base.cc
romange: 1. Fixed.
2. Each proactor has their own submission queue, so I do not thing starvation is possible. Having said that I added logs so we could learn if those events happen.

dranikpg (approved):

PR #255, opened 2024-04-21
fix: allow consistent traversals

romange:
Fixes https://github.com/dragonflydb/dragonfly/issues/2820

### DIFF
--- a/util/fibers/listener_interface.cc
+++ b/util/fibers/listener_interface.cc
@@ -36,7 +36,7 @@ using ListType =
 
 }  // namespace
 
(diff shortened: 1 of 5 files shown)


dranikpg (approved):
The lock is really tricky 😆 I understand that this is to avoid deadlocking like you can easily do with two counters - but the code is called rarely, a mutex would do.

romange:
The mutex would contend between migrations. 
We have users using 20k connections with bullmq workload and 10+ connection rate that migrate. You grab mutex for the whole time you create a bottleneck.

PR #259, opened 2024-05-04
feat: LinuxFile supports direct fd descriptors

romange:
Also fix a bug where we override kver.kernel inside uring proactor (affects kernels 5.x)

### DIFF
--- a/util/fibers/uring_file.h
+++ b/util/fibers/uring_file.h
@@ -40,9 +40,13 @@ class LinuxFile {
 
   std::error_code Close();
 
+  // Kept for backward compatibility. Use GetFd.
   int fd() const {
-    return fd_;
+    return GetFd();
   }
+
+  // Always returns linux file descriptor.
+  int GetFd() const;
 
   // Similar to XXXSome methods but writes fully all the io vectors or fails.
   // Is not atomic meaning that it can issue multiple io calls until error occurs or we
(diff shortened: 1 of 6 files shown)


util/fibers/uring_file.h
```diff
@@ -59,14 +63,24 @@ class LinuxFile {
...
   std::error_code ReadFixed(io::MutableBytes dest, off_t offset, unsigned buf_index);
 
+  // int - io result.
   using AsyncCb = std::function<void(int)>;
 
   // Async versions of Read
   void ReadFixedAsync(io::MutableBytes dest, off_t offset, unsigned buf_index, AsyncCb cb);
   void ReadAsync(io::MutableBytes dest, off_t offset, AsyncCb cb);
+  void WriteFixedAsync(io::Bytes src, off_t offset, unsigned buf_index, AsyncCb cb);
+  void WriteAsync(io::Bytes src, off_t offset, AsyncCb cb);
```
dranikpg, line 73: With those changes, can't we then just replace our IoMgr from DF with a plain LinuxFile?

dranikpg: Btw, what do you think about overloads accepting UringBuf right here?

dranikpg (approved):

util/fibers/uring_file.h
romange: > With those changes, can't we then just replace our IoMgr from DF with a plain LinuxFile?

That's the plan.

romange: > Btw, what do you think about overloads accepting UringBuf right here?

Why is it better?

dranikpg: > Why is it better?

Just more convenient, we use fixed reads only with uring buf

PR #261, opened 2024-05-05
chore: partial support of absl::log library

romange:
Keep the default with the old glog library. The absl log library is missing a LOT of features, like logging into files, file rotation etc. Basically it provides only the low-level api to serialize log messes into sink and a stderr sink. This is not enough for prod use. Still, this PR provides a common ground to use both libraries in neutral way. If I have bandwidth I will fill in the functionality inside helio to be able to get rid of old glog


romange:
@chakaz  not urgent, when you have time  :)

chakaz (dismissed):
I could swear that I already approved this, sorry!!

chakaz (approved):

PR #264, opened 2024-05-09
chore: Build mimalloc as a C++ library

chakaz:
This will have the benefit of throwing exceptions if allocations fail in the C++ APIs (like `new`)

Related to [#2914](https://github.com/dragonflydb/dragonfly/issues/2914)


cmake/third_party.cmake
```diff
@@ -297,9 +297,9 @@ add_third_party(mimalloc
    PATCH_COMMAND "${MIMALLOC_PATCH_COMMAND}"
    # -DCMAKE_BUILD_TYPE=Release
    # Add -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS=-O0 to debug
-   CMAKE_PASS_FLAGS "-DCMAKE_BUILD_TYPE=Release -DMI_BUILD_SHARED=OFF -DMI_BUILD_TESTS=OFF \
-                    -DMI_INSTALL_TOPLEVEL=ON -DMI_OVERRIDE=OFF -DCMAKE_C_FLAGS=-g \
-                    ${HELIO_MIMALLOC_OPTS}"
+   CMAKE_PASS_FLAGS "-DCMAKE_BUILD_TYPE=Release -DMI_USE_CXX=ON -DMI_BUILD_SHARED=OFF \
```
chakaz, line 300: I only added `-DMI_USE_CXX=ON`, the rest is just bad diff display :|

chakaz:
@romange do you think we should merge this PR?
I think that, given we know of scenarios in which mimalloc could return `nullptr` (even if not in our production, although we also suspect other cases in our production), it might be a good idea.
What are the downsides of doing so? It's the same code, if you compile it with a C or C++ compiler, that shouldn't matter I think

romange:
https://youtu.be/L0MK7qz13bU?si=ixfsy3fkBDjZ45MR

romange:
Closing for now :)

PR #268, opened 2024-05-17
feat: add reusable fibers and fiber pool

BorysTheDev:
It's just an idea. On my PC for test:
memtier_benchmark -t 8 --pipeline=30 --hide-histogram --test-time=3000 --distinct-client-seed -d 256 --ratio=1:0
I've got 19% performance improvement


romange:
Fibers are created once when the benchmark is starting and continue running during the lifetime of the loadtest program.
Based on this PR I do not see how these changes can affect the performance of the backend, specifically the performance of processing of requests.

BorysTheDev:
Fiber creation and destruction take a significant part of CPU in our code. In my PR, I create a pool that contains all unused fibers, and next time when we need a new fiber we just get already created and don't waste time.

PR #273, opened 2024-06-03
fix visibility of LEGACY_GLOG in cmake/third_party

kostasrim:
The problem was the visibility of option `LEGACY_GLOG` which was not available when `included` the thirdparty cmake.

### DIFF
--- a/cmake/third_party.cmake
+++ b/cmake/third_party.cmake
@@ -17,6 +17,7 @@ Include(ExternalProject)
 Include(FetchContent)
 
 option (WITH_UNWIND "Enable libunwind support" ON)
+option (LEGACY_GLOG "whether to use legacy glog library" ON)
 
 set(THIRD_PARTY_LIB_DIR "${THIRD_PARTY_DIR}/libs")
 file(MAKE_DIRECTORY ${THIRD_PARTY_LIB_DIR})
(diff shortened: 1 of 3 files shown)


cmake/third_party.cmake
```diff
@@ -17,6 +17,7 @@ Include(ExternalProject)
 Include(FetchContent)
 
 option (WITH_UNWIND "Enable libunwind support" ON)
+option (LEGACY_GLOG "whether to use legacy glog library" ON)
```
kostasrim, line 20: I tried to play around with variable visibility but I couldn't find a better way than duplicating the option. If there is a clever way to propagate this from the code that includes this I am all eyes :eyes:

romange: can you explain what the problem is?

romange: you can also remove `LEGACY_GLOG` from `CMakeLists.txt` and leave it only here

kostasrim: Yes, option `LEGACY_GLOG` is not seen by `third_party.cmake`. You define the variable/option in `CmakeLists` which includes `third_party.cmake`. The problem is that the code included, does not observe the variable `LEGACY_GLOG` so it doesn't fetch anytthing

kostasrim: > you can also remove LEGACY_GLOG from CMakeLists.txt and leave it only here 

Sounds good

romange: it is seen if you build helio separately. it is not seen in dragonfly. keeping it only here might fix the issue.

kostasrim: yes otherwise helio builds would have failed -- it makes sense. What `I wanted` to understand is why the variables are not visible. I did google it but I couldn't find something relevant and I also tried to use different scopes and replacing `OPTION` with `SET`. None of my attempts worked so I ended up to the fallback mechanism you see. 

P.s. I will remove the option from `CMakeLists` and it should be ok.

romange (approved):

PR #278, opened 2024-06-05
fix(tls): Fix spin_count checks

chakaz:


kostasrim:
@chakaz I think @romange specifically asked for `0` bytes? Double checking

romange (approved):

chakaz:
> @chakaz I think @romange specifically asked for `0` bytes? Double checking

Negative response still means zero bytes, but that's also an error.
Kostas, am I mistaken in saying that this is why we did not see the spin_count logs printed out?

PR #280, opened 2024-06-07
chore: Add a test for thread-local condvar.

romange:


dranikpg (approved):
not very strong but works🤷🏻‍♂️

PR #284, opened 2024-06-20
chore: add runtime detection of CPU extensions

romange:


chakaz (approved):

base/cxx_test.cc
```diff
@@ -190,4 +172,11 @@ TEST_F(CxxTest, Iterator) {
   EXPECT_EQ(sum, 6);
 }
 
+#ifdef __x86_64__
```
chakaz, line 175: I guess there's no reason to limit this test to x86..

chakaz (approved):

PR #288, opened 2024-07-09
fix: define macro WITH_AWS in cmake when flag is ON

kostasrim:
* fix missing preprocessor macro when WITH_AWS is on


cmake/third_party.cmake
```diff
@@ -27,7 +27,7 @@ file(MAKE_DIRECTORY ${THIRD_PARTY_LIB_DIR})
 set(THIRD_PARTY_CXX_FLAGS "-std=c++14 -O3 -DNDEBUG -fPIC -fno-stack-protector \
     -fno-stack-clash-protection")
 
-if (APPLE) 
+if (APPLE)
```
kostasrim, line 30: clang format ^^

kostasrim:
Closing this, moved it to DF directly

PR #300, opened 2024-07-25
fix: a better fix to sleeping fibers starvation

romange:
Before we used a counter which allowed inner loops to shortcut and rarely reach ProcessSleepFibers call in high I/O situations.

Now, we also call ProcessSleepFibers when we process I/O completions that provides a good balance between responsiveness and the maintenance cost of checking for starvation.

## Results:

Before I could observe delays reaching 2seconds between Heartbeat runs that where scheduled to run every 10ms (on highly loaded server in debug mode).

After: the delays have not crossed 40ms under the same conditions.

### DIFF
--- a/util/accept_server_test.cc
+++ b/util/accept_server_test.cc
@@ -276,8 +276,11 @@ TEST_F(AcceptServerTest, Shutdown) {
   constexpr auto kHupMask = POLLHUP;
 #endif
 
-    vector<unique_ptr<FiberSocketBase>> socks;
     constexpr unsigned kNumSocks = 2000;
+
+    vector<unique_ptr<FiberSocketBase>> socks;
+    vector<bool> error_notification(kNumSocks, false);
+
     unsigned called = 0;
     fb2::CondVarAny cond;
 
@@ -286,9 +289,14 @@ TEST_F(AcceptServerTest, Shutdown) {
       error_code ec = socks.back()->Connect(ep_);
       EXPECT_FALSE(ec);
 
-      socks.back()->RegisterOnErrorCb([&](int err) {
+      socks.back()->RegisterOnErrorCb([&, i](int err) {
         EXPECT_EQ(err, kHupMask);
-        ++called;
+
+        // MacOs can send multiple notifications per socket.
+        if (!error_notification[i]) {
+          ++called;
+          error_notification[i] = true;
+        }
         cond.notify_one();
       });
     }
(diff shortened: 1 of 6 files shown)


util/fibers/uring_proactor.cc
```diff
@@ -717,24 +716,23 @@ void UringProactor::MainLoop(detail::Scheduler* scheduler) {
...
-
     ///
     /// End of the tight loop that processes tasks, ready fibers, and submits sqes.
     ///
-    if (scheduler->HasSleepingFibers()) {
-      unsigned activated = ProcessSleepFibers(scheduler);
-      if (activated > 0) {  // If we have ready fibers - restart the loop.
-        continue;
-      }
+    unsigned activated = ProcessSleepFibers(scheduler);
```
romange, line 733: moved the HasSleepingFibers check into ProcessSleepFibers

util/fibers/uring_proactor.cc
```diff
@@ -717,24 +716,23 @@ void UringProactor::MainLoop(detail::Scheduler* scheduler) {
       // do now execute any syscalls. Therefore we count how many completions we handled
       // and reap the same amount.
       ReapCompletions(cqe_count, cqes, dispatcher);
+
+      if (HasSleepFibersStarved()) {
+        ProcessSleepFibers(scheduler);
```
chakaz, line 721: Since you're checking `HasSleepFibersStarved()` inside `ProcessSleepFibers()`, I'd remove the above check (or the one inside).

chakaz (approved):

util/fibers/uring_proactor.cc
romange: i do not check for HasSleepFibersStarved there

chakaz: Sorry, my bad

chakaz (dismissed):

romange:
Fixed the Waker code that did not work well on macos and also fixed accept server test

util/accept_server_test.cc
```diff
@@ -276,25 +276,36 @@ TEST_F(AcceptServerTest, Shutdown) {
...
 
-      socks.back()->RegisterOnErrorCb([&](int err) {
+      socks.back()->RegisterOnErrorCb([&, i](int err) {
         EXPECT_EQ(err, kHupMask);
-        ++called;
+
+        // MacOs can send multiple notifications per socket.
+        if (!error_notification[i]) {
+          ++called;
+          error_notification[i] = true;
```
chakaz, line 298: Unfortunately, this is unsafe (and it is un-intuitive that it is unsafe).
`vector<bool>` is defined in the standard to not be an array of `bool` but rather use a single bit per boolean. It's been considered an unfixable mistake.
https://en.cppreference.com/w/cpp/container/vector_bool
As such, accessing different indices from the different threads may yield undefined behavior...

A common fix is to use `vector<MyEnum>` or `vector<uint8_t>`
(I understand this is just a test, but still, let's make sure it doesn't fail due to this)

romange: but I do not access different indices from different threads.
It's all the same thread - it's all wrapped in one huge Await.

romange: it would be unsafe even with uint8_t if it were from different thrreads.

chakaz: > but I do not access different indices from different threads. It's all the same thread - it's all wrapped in one huge Await.

Right, I missed that.

> it would be unsafe even with uint8_t if it were from different thrreads.

That's true :)

chakaz (approved):

PR #301, opened 2024-07-27
chore: add ability to print custom data during PrintAllFiberStackTraces

romange:
This can be used to debug deadlock states when the state is kept within the fiber stack.


romange:
Motivation: https://github.com/dragonflydb/dragonfly/actions/runs/10123280427/job/27996377860?pr=3384#step:10:1820

dranikpg (approved):

PR #304, opened 2024-08-11
chore: allow checking if Fiber is active

romange:


chakaz (approved):
LG. We could add a unit test for that, just saying :nerd_face:

PR #305, opened 2024-08-12
chore: thread safety annotations on synchronization primitives

kostasrim:
In order for our mutex classes to work with [clang's ThreadSafety analysis](https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#thread-safety-analysis) (used by ABSL_GUARD_XXX functions) we need to properly annotate them. This PR adds the missing annotations.

### DIFF
--- a/base/spinlock.h
+++ b/base/spinlock.h
@@ -8,21 +8,21 @@
 
 namespace base {
 
-class SpinLock {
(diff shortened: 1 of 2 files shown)


kostasrim:
@romange this is not ready, I will ping

romange (approved):

PR #311, opened 2024-09-02
fix: reset kScheduleRemote when we pull the queue

romange:
Before, we reset kScheduleRemote bit right after we pushed into the remote queue.
    But if this thread is slow, it is possible that the notified fiber will wake up before we reset the bit,
    and then it will go sleeping again and then another notifier will try to awake it and will check fail at line
    scheduler.cc:238 when it checks that the bit is set.
    If we reset it upon waking up, the fiber will reset it before suspending, so this situation can not happen.
    
    In addition, we print the stacktrace when we reach unexpected state.

### DIFF
--- a/util/fibers/detail/scheduler.cc
+++ b/util/fibers/detail/scheduler.cc
@@ -235,7 +235,7 @@ void Scheduler::ScheduleFromRemote(FiberInterface* cntx) {
   // This should not happen as ScheduleFromRemote should be called under a WaitQueue lock.
   if ((cntx->flags_.fetch_or(FiberInterface::kScheduleRemote, memory_order_acquire) &
        FiberInterface::kScheduleRemote) != 0) {
-    LOG(DFATAL) << "Already scheduled remotely " << cntx->name();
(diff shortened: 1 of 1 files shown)


util/fibers/detail/scheduler.cc
```diff
@@ -253,10 +253,6 @@ void Scheduler::ScheduleFromRemote(FiberInterface* cntx) {
     cntx->DEBUG_remote_epoch = remote_epoch_.fetch_add(1, memory_order_relaxed);
     remote_ready_queue_.Push(cntx);
 
-
-    // clear the bit after we pushed to the queue.
-    cntx->flags_.fetch_and(~FiberInterface::kScheduleRemote, memory_order_release);
```
kostasrim, line 258: I find it slightly confusing that if `isScheduledRemotely()` is true above we clear the bit here whereas for this case we clear the bit when we pop the item in from the remote queue in the remote scheduler. What I had in mind is to clear the bit here and on the remote scheduler side to place back the item at the back of the remote_queue. That way this flow still remains in control of unseting the bit. 

IMO I don't see a clear benefit in my suggestion so your change good as well.

kostasrim (approved):

chakaz (approved):

PR #312, opened 2024-09-04
fix: SegmentPool when we allocate more than capacity

romange:


dranikpg (approved):
Silly bug from my side 🙈

PR #316, opened 2024-09-09
chore: add more vlog info in RunAcceptLoop

kostasrim:
Add more logs to figure out where we deadlock during shutdown in df


util/fibers/listener_interface.cc
```diff
@@ -193,10 +196,12 @@ void ListenerInterface::RunAcceptLoop() {
       if (guard->hook_.is_linked()) {
         ++iter;
       } else {
+        VLOG(1) << "Iterator reset for pb with thread_id " << pb->thread_id();
```
kostasrim, line 199: This `else` loop could lead us to a never ending `while()` loop. Let's see if the following logs provide us with more info. Also Roman, if you have anything to add here I am all eyes :eyes:

romange (approved):
lgtm

PR #344, opened 2024-11-23
chore: improve signal handling code

romange:
1. Relax ClearSignal function - now it can be called on unregistered signals (before it check-failed)
2. Now it is possible to install the ignore handler,
   which is what we really want to use when we clear the signal handlers via AcceptServer.
   Indeed, if accept server is shut down, the process will usually exit right after.
   There will be some clean up code but that's it. So we do not want to interrupt due sigterm/sigint signals because
   we are already in the process of shutting down anyway. Before this change - if we got the signal during the
   critical section after the accept server handlers were cleared, and default handlers were installed,
   the process would exit with -15 for SIGTERM and -2 for SIGINT just because it would catch these signals
   during the shutdown.

### DIFF
--- a/util/fibers/accept_server.cc
+++ b/util/fibers/accept_server.cc
@@ -33,7 +33,7 @@ AcceptServer::AcceptServer(ProactorPool* pool, PMR_NS::memory_resource* mr, bool
 
 AcceptServer::~AcceptServer() {
   if (break_on_int_) {
-    ProactorBase::ClearSignal({SIGINT, SIGTERM});
+    ProactorBase::ClearSignal({SIGINT, SIGTERM}, true);
   }
   list_interface_.clear();
 }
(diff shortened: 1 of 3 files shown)


util/fibers/accept_server.cc
```diff
@@ -33,7 +33,7 @@ AcceptServer::AcceptServer(ProactorPool* pool, PMR_NS::memory_resource* mr, bool
 
 AcceptServer::~AcceptServer() {
   if (break_on_int_) {
-    ProactorBase::ClearSignal({SIGINT, SIGTERM});
+    ProactorBase::ClearSignal({SIGINT, SIGTERM}, true);
```
chakaz, line 36: You already know what I think about boolean arguments :rofl:

chakaz (approved):

util/fibers/accept_server.cc
romange: next time i fix something here, I will change

PR #363, opened 2025-01-09
fix: EpollSocket::WriteSome early return on EAGAIN

kostasrim:
`sendmsg` returns the bytes written and `-1` if it failed. For the later, we need to poll ERNO, and for `EAGAIN` we need to block and retry later when we get notified. What this means is that the calling fiber must block on EAGAIN when we call WriteSome. However this is not the case as we improperly check: `if (res != EAGAIN)` which can't be true because we are on the path of `sendmsg` that returned `< 0`. We should check `ERRNO` and not `res`.

### DIFF
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -4,6 +4,7 @@
 
 #include "util/fibers/epoll_socket.h"
 
+#include <errno.h>
 #include <netinet/in.h>
 
 #include "absl/cleanup/cleanup.h"
@@ -134,7 +135,6 @@ class EpollSocket::PendingReq {
 
   void Activate(error_code ec);
 };
-
 
 error_code EpollSocket::PendingReq::Suspend(uint32_t timeout) {
   bool timed_out = false;
@@ -352,7 +352,7 @@ auto EpollSocket::WriteSome(const iovec* ptr, uint32_t len) -> Result<size_t> {
 
     DCHECK_EQ(res, -1);
 
-    if (res != EAGAIN) {
+    if (errno != EAGAIN) {
       ec = from_errno();
       break;
     }
--- a/util/fibers/fiber_socket_test.cc
+++ b/util/fibers/fiber_socket_test.cc
@@ -167,12 +167,6 @@ TEST_P(FiberSocketTest, Basic) {
 }
 
 TEST_P(FiberSocketTest, Bug363) {
-  const bool use_epoll = GetParam() == "epoll";
-  // We test only epoll proactor
-  if (!use_epoll) {
-    return;
-  }
-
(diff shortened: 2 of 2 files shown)


util/fibers/epoll_socket.cc
```diff
@@ -352,7 +352,7 @@ auto EpollSocket::WriteSome(const iovec* ptr, uint32_t len) -> Result<size_t> {
 
     DCHECK_EQ(res, -1);
 
-    if (res != EAGAIN) {
+    if (errno != EAGAIN) {
```
kostasrim, line 355: You even `DCHECK_EQ(res, -1);` above :rofl:

romange: ouch.

romange (approved):

util/fibers/epoll_socket.cc
kostasrim: You have it right in other places I checked :smile:

romange:
I wish you also add a test trigerring this path. We have fiber_socket_test but clearly it did not cover this.

kostasrim:
> I wish you also add a test trigerring this path. We have fiber_socket_test but clearly it did not cover this.

Happy to do that. Don't merge this, I will push a commit with a test. 

P.s. won't do it right now, have my attention over other stuff atm

util/fibers/fiber_socket_test.cc
```diff
@@ -160,11 +160,41 @@ TEST_P(FiberSocketTest, Basic) {
...
     auto res = sock->WriteSome(io::Bytes(buf));
     EXPECT_EQ(16, res.value_or(0));
+
     VLOG(1) << "closing client sock " << sock->native_handle();
     std::ignore = sock->Close();
   });
 }
 
+TEST_P(FiberSocketTest, Bug363) {
+  const bool use_epoll = GetParam() == "epoll";
```
romange, line 170: no reason to limit it to Epoll

util/fibers/fiber_socket_test.cc
```diff
@@ -160,11 +160,41 @@ TEST_P(FiberSocketTest, Basic) {
...
+    std::ignore = sock->Connect(listen_ep_);
+    accept_fb_.Join();
+
+    // Explode the socket with Writes until it blocks. At this point the socket
+    // should timeout;
+    for (int i = 0; i < 50; ++i) {
+      uint8_t buf[1024 * 100] = {0};
+      auto res = sock->WriteSome(io::Bytes(buf));
+      if (!res) {
+        ASSERT_TRUE(res.error().value() != EAGAIN);
```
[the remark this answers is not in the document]
kostasrim: we return `ECANCELED`. Fixed :)

romange (approved):

PR #364, opened 2025-01-11
fix: make GCS authentication honor Application Default Credentials

nmdanny:
ADC will now be retrieved as follows:

1. If `GOOGLE_APPLICATION_CREDENTIALS` exists, it will be used to determine the ADC file (new)

2. Otherwise, if ~/.config/gcloud exists, get the credentials from the active account.

3. Otherwise, fetch from metadata server

Fixes [dragonflydb/dragonfly/#4396](https://github.com/dragonflydb/dragonfly/issues/4396)

### DIFF
--- a/util/cloud/gcp/gcs.cc
+++ b/util/cloud/gcp/gcs.cc
@@ -35,6 +35,11 @@ auto Unexpected(std::errc code) {
 
 const char kInstanceTokenUrl[] = "/computeMetadata/v1/instance/service-accounts/default/token";
 
+const char kGcloudDir[] = "~/.config/gcloud/";
+const char kGcloudDefaultConfigPath[] = "configurations/config_default";
+const char kGcloudCredentialsFolder[] = "legacy_credentials/";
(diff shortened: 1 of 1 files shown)


util/cloud/gcp/gcs.cc
```diff
@@ -35,6 +35,11 @@ auto Unexpected(std::errc code) {
 
 const char kInstanceTokenUrl[] = "/computeMetadata/v1/instance/service-accounts/default/token";
 
+const char gcloudCliDir[] = "~/.config/gcloud/";
```
romange, line 38: notation for constants is `kCloudCliDir`. here and below

romange (commented):
thanks! looks good.
a small style comment and we are good to merge

util/cloud/gcp/gcs.cc
nmdanny: IMO `kGcloudCliDir` makes more sense (emphasis on `gcloud` since it's specifically google's CLI, not something generic)

romange (approved):

PR #367, opened 2025-01-13
feat(fibers): Add GetStackMargin method

BagritsevichStepan:
During debugging, it is sometimes useful to check how much stack space is left before calling a method. To do this, we can use `GetStackMargin`.

### DIFF
--- a/util/fibers/fibers.h
+++ b/util/fibers/fibers.h
@@ -175,14 +175,19 @@ inline std::string_view GetName() {
   return fb2::detail::FiberActive()->name();
 }
 
+/* Returns the margin between the provided stack address
+   and the bottom of the fiber's stack. */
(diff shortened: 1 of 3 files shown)


util/fibers/fibers.h
```diff
@@ -175,14 +175,17 @@ inline std::string_view GetName() {
   return fb2::detail::FiberActive()->name();
 }
 
+inline uint32_t GetSafetyMargin() {
```
romange, line 178: please add a comment what it returns and make sure both function names are consistent.

BagritsevichStepan: Ah, yeah, it's a typo

romange:
@BagritsevichStepan  please sign your commit.

romange (approved):

PR #370, opened 2025-01-15
chore: update sse2neon.h to 01/2025 version

romange:
The files is taken from https://github.com/DLTcollab/sse2neon and it uses the MIT license.


kostasrim:
I will check it out with sanitizers once I get a chance

PR #374, opened 2025-01-23
fix: epoll socket async chaining and refactor

kostasrim:
Fixes the problem described in https://github.com/dragonflydb/dragonfly/issues/4498 and here https://github.com/dragonflydb/dragonfly/issues/4498#issuecomment-2609240866

* fix async chaining
* small refactor

### DIFF
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -567,6 +567,41 @@ void EpollSocket::CancelOnErrorCb() {
   error_cb_ = {};
 }
 
+void EpollSocket::HandleAsyncRequest(error_code ec, bool is_send) {
+  auto async_pending = is_send ? async_write_pending_ : async_read_pending_;
+  if (async_pending) {
+    auto& async_request = is_send ? async_write_req_ : async_read_req_;
+    DCHECK(async_request);
+
+    auto finalize_and_fetch_cb = [this, &async_request, is_send]() {
+      auto cb = std::move(async_request->cb);
+      delete async_request;
+      async_request = nullptr;
+      if (is_send)
+        async_write_pending_ = 0;
+      else
+        async_read_pending_ = 0;
+      return cb;
+    };
+
+    if (ec) {
+      auto cb = finalize_and_fetch_cb();
+      cb(make_unexpected(ec));
+    } else if (auto res = async_request->Run(native_handle(), is_send); res.first) {
+      auto cb = finalize_and_fetch_cb();
+      cb(res.second);
+    }
+  } else {
+    auto& sync_request = is_send ? write_req_ : read_req_;
+    // It could be that we activated context already, but has not switched to it yet.
+    // Meanwhile a new event has arrived that triggered this callback again.
+    if (sync_request && sync_request->IsSuspended()) {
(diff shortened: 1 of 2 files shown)


kostasrim:
@romange Fixes the test case as well in DF. I could add a test but that's something I will be doing anyway when I push my `AsyncWrite/AsyncRead` on the tls socket.

util/fibers/epoll_socket.cc
```diff
@@ -583,55 +582,49 @@ void EpollSocket::Wakey(uint32_t ev_mask, int error, EpollProactor* cntr) {
...
+    if (async_pending) {
+      auto& async_request = is_send ? async_write_req_ : async_read_req_;
+      DCHECK(async_request);
+
+      auto finalize_and_fetch_cb = [this, &async_request, is_send]() {
+        auto cb = std::move(async_request->cb);
+        delete async_request;
+        async_request = nullptr;
+        if (is_send)
+          async_write_pending_ = 0;
```
kostasrim, line 596: Cause lambdas can't capture and use bitfields :face_exhaling:

util/fibers/epoll_socket.cc
```diff
@@ -583,55 +582,49 @@ void EpollSocket::Wakey(uint32_t ev_mask, int error, EpollProactor* cntr) {
...
 
-  if (ev_mask & (EpollProactor::EPOLL_IN | kErrMask)) {
-    if (async_read_pending_) {
-      DCHECK(async_read_req_);
-
-      auto finalize = [this] {
-        delete async_read_req_;
-        async_read_req_ = nullptr;
-        async_read_pending_ = 0;
+  auto body = [&ec, this](bool is_send) {
```
romange, line 585: since it captures `this`, please make it a member function that returns `error_code`.

util/fibers/epoll_socket.cc
```diff
@@ -166,22 +166,18 @@ bool EpollSocket::AsyncReq::Run(int fd, bool is_send) {
   res = is_send ? sendmsg(fd, &msg, MSG_NOSIGNAL) : recvmsg(fd, &msg, 0);
 
   if (res > 0) {
-    cb(res);
-    return true;
+    return {true, res};
```
romange, line 169: what is the reason for moving cb() outside of AsyncReq::Run?

util/fibers/epoll_socket.cc
kostasrim: +1

util/fibers/epoll_socket.cc
kostasrim: This is the fix. If `cb()` calls again `AsyncWrite` it will `CHECK FAIL` because the `AsyncReq` is not `nullptr`. We have to reset the `async request` before we call the callback and to do that, we need to move it outside of the object we are about to deallocate

util/fibers/epoll_socket.cc
kostasrim: So much better now, I really don't know what I was thinking with this large lambda

kostasrim:
@romange there seem to be a problem with docker on the CI. I also can't rerun the workflow :cry:

romange (approved):

PR #376, opened 2025-01-31
[do not review] feat: TlsSocket AsyncWriteSome and AsyncReadSome

kostasrim:
Add `AsyncWriteSome` and `AsyncReadSome` to `TlsSocket`.

There are a few things that need to be fixed/implemented:

1. To "yield" when there a pending read or write (to avoid deadlock) see HandleOp comments
2. To clean up the state such that progress callbacks can re-request an async operation
3. Clean up the main loop of AsyncWrite
4. Improve testing and also force tls protocol renegotiation
5. Generate tls certificates for CI testing
6. Investigate iouring imemory leak

### DIFF
--- a/util/tls/tls_socket.cc
+++ b/util/tls/tls_socket.cc
@@ -329,21 +329,215 @@ io::Result<TlsSocket::PushResult> TlsSocket::PushToEngine(const iovec* ptr, uint
   return res;
 }
 
-// TODO: to implement async functionality.
-void TlsSocket::AsyncWriteSome(const iovec* v, uint32_t len, io::AsyncProgressCb cb) {
-  io::Result<size_t> res = WriteSome(v, len);
-  cb(res);
+void TlsSocket::HandleOpAsync(int op_val) {
+  switch (op_val) {
+    case Engine::EOF_STREAM:
+      VLOG(1) << "EOF_STREAM received " << next_sock_->native_handle();
+      async_write_req_->caller_completion_cb(
+          make_unexpected(make_error_code(errc::connection_aborted)));
+      break;
+    case Engine::NEED_READ_AND_MAYBE_WRITE:
+      HandleUpstreamAsyncRead();
+      break;
+    case Engine::NEED_WRITE:
+      MaybeSendOutputAsync();
+      break;
+    default:
+      LOG(DFATAL) << "Unsupported " << op_val;
+  }
 }
 
+void TlsSocket::AsyncWriteReq::Run() {
+  if (state == AsyncWriteReq::PushToEngine) {
+    io::Result<PushResult> push_res = owner->PushToEngine(vec, len);
+    if (!push_res) {
+      caller_completion_cb(make_unexpected(push_res.error()));
+      return;
+    }
+    last_push = *push_res;
+    state = AsyncWriteReq::HandleOpAsync;
+  }
 
-// TODO: to implement async functionality.
+  if (state == AsyncWriteReq::HandleOpAsync) {
+    state = AsyncWriteReq::MaybeSendOutputAsync;
+    if (last_push.engine_opcode < 0) {
+      owner->HandleOpAsync(last_push.engine_opcode);
+    }
+  }
+
+  if (state == AsyncWriteReq::MaybeSendOutputAsync) {
+    state = AsyncWriteReq::PushToEngine;
+    if (last_push.written > 0) {
+      DCHECK(!continuation);
+      continuation = [this]() {
+        state = AsyncWriteReq::Done;
+        caller_completion_cb(last_push.written);
+      };
+      owner->MaybeSendOutputAsync();
+    }
+  }
+}
+
+void TlsSocket::AsyncWriteSome(const iovec* v, uint32_t len, io::AsyncProgressCb cb) {
+  CHECK(!async_write_req_.has_value());
+  async_write_req_.emplace(AsyncWriteReq(this, std::move(cb), v, len));
+  async_write_req_->Run();
+}
+
+void TlsSocket::AsyncReadReq::Run() {
+  DCHECK_GT(len, 0u);
+
+  while (true) {
+    DCHECK(!dest.empty());
+
+    size_t read_len = std::min(dest.size(), size_t(INT_MAX));
+
+    Engine::OpResult op_result = owner->engine_->Read(dest.data(), read_len);
+
+    int op_val = op_result;
+
+    DVLOG(2) << "Engine::Read " << dest.size() << " bytes, got " << op_val;
+
+    if (op_val > 0) {
+      read_total += op_val;
+
+      // I do not understand this code and what the hell I meant to do here. Seems to work
+      // though.
+      if (size_t(op_val) == read_len) {
+        if (size_t(op_val) < dest.size()) {
+          dest.remove_prefix(op_val);
+        } else {
+          ++vec;
+          --len;
+          if (len == 0) {
+            // We are done. Call completion callback.
+            caller_completion_cb(read_total);
+            return;
+          }
+          dest = Engine::MutableBuffer{reinterpret_cast<uint8_t*>(vec->iov_base), vec->iov_len};
+        }
+        // We read everything we asked for but there are still buffers left to fill.
+        continue;
+      }
+      break;
+    }
+
+    // Will automatically call Run()
+    owner->HandleOpAsync(op_val);
+  }
+
+  // We are done. Call completion callback.
+  caller_completion_cb(read_total);
+
+  // clean up so we can queue more reads
+}
+
 void TlsSocket::AsyncReadSome(const iovec* v, uint32_t len, io::AsyncProgressCb cb) {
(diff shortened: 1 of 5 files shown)


kostasrim:
Do not review, it's not ready

util/tls/certificates/ca-cert.pem
```diff
@@ -0,0 +1,34 @@
+-----BEGIN CERTIFICATE-----
```
kostasrim, line 1: I think `50 years` will be enough for this to last :nerd_face:

io/io.cc
```diff
@@ -50,6 +50,7 @@ struct AsyncReadState {
 
   AsyncReadState(AsyncSource* source, const iovec* v, uint32_t length)
       : arr(length), owner(source) {
+    cur = arr.data();
```
kostasrim, line 53: Ouch -- we not got unit tests though :smile:

util/tls/certificates/ca-cert.pem
kostasrim: IMO I realized that we can generate the certificates with cmake as well. Not sure if it's worth it or not :thinking:

vyavdoshenko: In my opinion, the better idea is to add a custom step for generating all keys and certificates on the cmake step.
You never know who or how can use these certificates and blame authors for that.

util/tls/tls_socket.h
```diff
@@ -114,6 +114,74 @@ class TlsSocket final : public FiberSocketBase {
...
+
+  struct AsyncReadReq : AsyncReqBase {
+    using AsyncReqBase::AsyncReqBase;
+
+    Engine::MutableBuffer dest;
+    size_t read_total = 0;
+
+    // Main loop
+    void Run() override;
+    virtual void CompleteAsyncReq(io::Result<size_t> result) override;
```
vyavdoshenko, line 175: override always requires to override virtual method.
virtual keyword is optional in such cases.

util/tls/certificates/ca-cert.pem
kostasrim: Good catch! I will fix. I also need to address a few things functionality/testing wise :)

util/tls/tls_socket.h
kostasrim: Good catch! I will fix. I also need to address a few things functionality/testing wise :)

romange:
Can I review it @kostasrim ?

util/tls/tls_socket.h
```diff
@@ -114,6 +114,80 @@ class TlsSocket final : public FiberSocketBase {
...
+    enum State { PushToEngine, HandleOpAsyncTag, MaybeSendOutputAsyncTag, Done };
+    State state = PushToEngine;
+    PushResult last_push;
+
+    // Main loop
+    void Run() override;
+    void CompleteAsyncReq(io::Result<size_t> result) override;
+  };
+
+  friend AsyncWriteReq;
```
romange, line 170: i think it will work without `friend` decl

util/tls/tls_socket.h
```diff
@@ -114,6 +114,80 @@ class TlsSocket final : public FiberSocketBase {
...
+    // Main loop
+    void Run() override;
+    void CompleteAsyncReq(io::Result<size_t> result) override;
+  };
+
+  friend AsyncReadReq;
+
+  // TODO clean up the optional before we yield such that progress callback can dispatch another
+  // async operation
+  std::optional<AsyncWriteReq> async_write_req_;
```
romange, line 187: most of our use with tls is synchronous, so I suggest to make them at least unique_ptr (could even be plain pointers), to reduce the size of TlsSocket for most cases.

this will also allow to move Async state definitions to cc file.

util/tls/tls_socket.cc
```diff
@@ -331,22 +331,254 @@ io::Result<TlsSocket::PushResult> TlsSocket::PushToEngine(const iovec* ptr, uint
...
+
+  // We are done. Call completion callback.
+  CompleteAsyncReq(read_total);
 }
 
-// TODO: to implement async functionality.
 void TlsSocket::AsyncReadSome(const iovec* v, uint32_t len, io::AsyncProgressCb cb) {
-  io::Result<size_t> res = ReadSome(v, len);
-  cb(res);
+  CHECK(!async_read_req_.has_value());
```
romange, line 453: This is not how I would design it. 
TlsEngine could already have a ready read buffer. you could return quickly by fetching it without doing any async stuff. Only if the buffer is empty it's worth dispatching the upstream read.

util/tls/tls_socket.cc
```diff
@@ -331,22 +331,254 @@ io::Result<TlsSocket::PushResult> TlsSocket::PushToEngine(const iovec* ptr, uint
...
+      break;
+    case Engine::NEED_WRITE:
+      MaybeSendOutputAsync();
+      break;
+    default:
+      LOG(DFATAL) << "Unsupported " << op_val;
+  }
+}
+
+void TlsSocket::AsyncWriteReq::Run() {
```
romange, line 361: why do we have `while (true)` here but we return in most cases?
it's hard to understand this logic - at first I thought Run is fiber-blocking.

util/tls/tls_socket.cc
```diff
@@ -331,22 +331,254 @@ io::Result<TlsSocket::PushResult> TlsSocket::PushToEngine(const iovec* ptr, uint
...
+      // We never preempt here
+      io::Result<PushResult> push_res = owner->PushToEngine(vec, len);
+      if (!push_res) {
+        CompleteAsyncReq(make_unexpected(push_res.error()));
+        // We are done with this AsyncWriteReq. Caller might started
+        // a new one.
+        return;
+      }
+      last_push = *push_res;
+      state = AsyncWriteReq::HandleOpAsyncTag;
```
romange, line 373: seems like switch case without breaks would work here

util/tls/tls_socket.cc
romange: what I am trying to understand - what is the path to the second loop?

util/tls/tls_socket.cc
```diff
@@ -331,22 +331,254 @@ io::Result<TlsSocket::PushResult> TlsSocket::PushToEngine(const iovec* ptr, uint
...
-  cb(res);
+  CHECK(!async_write_req_.has_value());
+  async_write_req_.emplace(AsyncWriteReq(this, std::move(cb), v, len));
+  async_write_req_->Run();
+}
+
+void TlsSocket::AsyncReadReq::Run() {
+  DCHECK_GT(len, 0u);
+
+  while (true) {
```
romange, line 408: here as well, why do we need a loop?

util/tls/tls_socket.h
kostasrim: Yep that makes sense. It's 130 bytes roughly extra per object

util/tls/tls_socket.cc
kostasrim: Well, isn't that what I do here? We don't start/submit any async operation. We construct
the object and call `Run()` and within this function we will read the EngineBuffer first if there are bytes available -- so we are on the fast path already, that is, we don't dispatch upstream.
As I am changing it now to a unique ptr, we now always get an allocation. I think the cost would be insignificant because the allocator will be able to satisfy that without actually requesting more memory and this is not a common path anyways.

util/tls/tls_socket.cc
kostasrim: There are no fiber blocks, everything is async and yes we need that. An example would be:

1. We push to the engine
2. We call HandleOpAsync because last_push.engine_opcode < 0 and we submit an async operation to the underline socket. Its completion handler will call `Run()` again.
3. The fiber continues doing its work while the underline async operation is in flight
4. The async operation completes, invokes the completion handler which calls `Run()` again. Our state now is `MaybeSendOutputAsyncTag` so we reach that branch but the condition `last_push.written() > 0` is not true. Now we loop, push to the engine again and proceed.

util/tls/tls_socket.cc
kostasrim: I need to check this one, I think it's redundant

kostasrim: I need to check this one

kostasrim:
> In general this PR is fairly big and it is hard to reason about design choices at this point. And also do a proper code review.
> 
> I suggest to start smaller. Say, implement only the read path, while use synchronous stubs to handle `Engine::NEED_WRITE` states. (I would even expect that for simple hello worlds scenarios, NEED_WRITE state won't even happen.)

I followed a similar design to what you did with async reads/writes for plain sockets. The difference is that the state machine for the tls stuff is a little bit more complicated and I find it harder to express those branches/flows async. For example, if we had coros, I would not need to do state tracking on some of the loops because once the completion handler for the inflight async operation completes, it would just call the coro handler which would resume from where it was suspended before the async operation started. Now it has to reach Run() first and drive down to its path. continuations would also help here, and I even took that route on a certain path (ofc not proper proper continuation).

PR #384, opened 2025-02-24
chore: gcp code bugs

romange:
1. Invalid release up of the pool client handle.
2. add a missing header for gcp metadata request.

### DIFF
--- a/util/cloud/gcp/gcs.cc
+++ b/util/cloud/gcp/gcs.cc
@@ -144,9 +144,11 @@ io::Result<TokenTtl> ParseTokenResponse(std::string&& response) {
   return result;
 }
 
+static constexpr string_view kMetaDataHost = "metadata.google.internal"sv;
+
 error_code ConfigureMetadataClient(fb2::ProactorBase* pb, http::Client* client) {
   client->set_connect_timeout_ms(1000);
-  static const char kMetaDataHost[] = "metadata.google.internal";
+
   return client->Connect(kMetaDataHost, "80");
 }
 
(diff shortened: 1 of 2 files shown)


util/cloud/gcp/gcs.cc
```diff
@@ -144,9 +144,11 @@ io::Result<TokenTtl> ParseTokenResponse(std::string&& response) {
   return result;
 }
 
+static const char kMetaDataHost[] = "metadata.google.internal";
```
chakaz, line 147:
```suggestion
static constexpr string_view kMetaDataHost = "metadata.google.internal"sv;
```
Just a nit, doesn't really matter

chakaz (approved):
LGTM but I'm not sure which headers are fixed by this? (from PR desc)

util/cloud/gcp/gcs.cc
```diff
@@ -158,6 +160,7 @@ error_code ReadGCPConfigFromMetadata(fb2::ProactorBase* pb, string* account_id,
   const char kEmailUrl[] = "/computeMetadata/v1/instance/service-accounts/default/email";
   h2::request<h2::empty_body> req{h2::verb::get, kEmailUrl, 11};
   req.set("Metadata-Flavor", "Google");
+  req.set(h2::field::host, kMetaDataHost);
```
romange, line 163: @chakaz  these headers :) host was missing so the requests were rejected

chakaz: oh lol, I was thinking of C++ headers :facepalm:

romange: fixed the PR description

chakaz (approved):
Now that I see that `req.set()` accepts a `const char*`, I had to verify that `string_view` literals ("..."sv) are null-terminated.
I looks like it is:
https://stackoverflow.com/questions/57250440/is-a-stdstring-view-literal-guaranteed-to-be-null-terminated
But still, it seems to have been a not-so-great advice on my behalf :|

Anyway, LG, do whatever you prefer

PR #387, opened 2025-03-02
chore: add zstd static library to helio dependencies

romange:


chakaz (approved):
Always better to build locally if possible & easy :)

romange:
yep, static linkage rulez :)

PR #388, opened 2025-03-02
chore: more efficient access to FiberInterface epoch field

romange:
To improve efficiency of fiber context switch operations.

### DIFF
--- a/util/fibers/detail/fiber_interface.h
+++ b/util/fibers/detail/fiber_interface.h
@@ -223,6 +223,9 @@ class FiberInterface {
 
   size_t GetStackMargin(const void* stack_address) const;
 
+  static uint64_t TL_Epoch() {
+    return tl.epoch;
+  }
  protected:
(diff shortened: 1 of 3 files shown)


util/fibers/detail/fiber_interface.h
```diff
@@ -262,6 +265,10 @@ class FiberInterface {
   uint32_t stack_size_ = 0;
   uint8_t* stack_bottom_ = nullptr;
 
+  static __thread struct TL {
+    uint64_t epoch = 0;
+  } tl;
```
chakaz, line 270: Why do you use a `struct` here instead of just `static __thread uint64_t epoch_ = 0;`?

chakaz: oh, gotcha, this is intended to (potentially) grow..

chakaz (approved):

romange:
Thanks! :)

PR #389, opened 2025-03-05
more: print more context if we preempt in atomic section

romange:


util/fibers/detail/scheduler.cc
```diff
@@ -203,8 +203,17 @@ ctx::fiber_context Scheduler::Preempt() {
...
     LOG(DFATAL) << "Should not preempt dispatcher: " << GetStacktrace();
   }
-  if (IsFiberAtomicSection())
-    LOG(DFATAL) << "Preempting inside of atomic section";
+
+  if (IsFiberAtomicSection()) {
+    static int64_t last_ts = 0;
+    int64_t now = time(nullptr);
+    if (now != last_ts) {  //   once a second at most.
+      last_ts = now;
```
chakaz, line 211: When you remove the `LEGACY_GLOG` support in favor of absl logs, you can replace this with a simple `LOG_EVERY_N_SEC(DFATAL, 1)`, but I guess not yet :disappointed:

chakaz:
> Out of context here

I see what you did there! :rofl:

util/fibers/detail/scheduler.cc
romange: yeah this does not work and I won't have time to complete all the logging functionality with absl. 
they do not support log files out of the box and other stuff as well. Could be a nice feature contribution to helio ;)

PR #390, opened 2025-03-12
chore: add debug printing for the Unlink bug

romange:
If for some reasone a fiber was resumed not via wait_list_, we identify this and print the error message. We also attempt to do the self-healing fix by unlinking the waiter manually.


romange:
Also run https://github.com/dragonflydb/dragonfly/actions/runs/13808681623 with this branch

PR #391, opened 2025-03-12
fix: ipv6 support updated

vyavdoshenko:
fix: ipv6 support updated
docs: The original issue is here: https://github.com/dragonflydb/dragonfly/issues/4622

### DIFF
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -267,7 +267,7 @@ error_code EpollSocket::Connect(const endpoint_type& ep, std::function<void(int)
 
   error_code ec;
 
-  int fd = ep.address().is_v4() ? CreateSockFd(AF_INET) : CreateSockFd(AF_INET6);
+  int fd = CreateSockFd(ep.address().is_v4() ? AF_INET : AF_INET6);
   if (posix_err_wrap(fd, &ec) < 0)
     return ec;
 
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -49,15 +49,20 @@ nonstd::unexpected<error_code> MakeUnexpected(std::errc code) {
 #ifdef __linux__
 constexpr int kEventMask = EPOLLIN | EPOLLOUT | EPOLLET | EPOLLRDHUP;
 
-int AcceptSock(int fd) {
-  sockaddr_in client_addr;
-  socklen_t addr_len = sizeof(client_addr);
-  int res = accept4(fd, (struct sockaddr*)&client_addr, &addr_len, SOCK_NONBLOCK | SOCK_CLOEXEC);
-  return res;
+int AcceptSock(int fd, bool is_v4) {
+  if (is_v4) {
+    sockaddr_in client_addr;
+    socklen_t addr_len = sizeof(client_addr);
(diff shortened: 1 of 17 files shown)


util/fibers/epoll_socket.cc
```diff
@@ -50,14 +50,13 @@ nonstd::unexpected<error_code> MakeUnexpected(std::errc code) {
 constexpr int kEventMask = EPOLLIN | EPOLLOUT | EPOLLET | EPOLLRDHUP;
 
 int AcceptSock(int fd) {
-  sockaddr_in client_addr;
+  sockaddr_storage client_addr;
```
romange, line 53: nice, I did not know about this type

vyavdoshenko: This type doesn't fit all the cases. But here it suits well.

util/fibers/fiber_socket_base.cc
```diff
@@ -95,15 +95,27 @@ error_code LinuxSocketBase::Create(unsigned short pfamily) {
...
-  sockaddr_in server_addr;
-  memset(&server_addr, 0, sizeof(server_addr));
-  server_addr.sin_family = AF_INET;
-  server_addr.sin_port = htons(port);
-  server_addr.sin_addr.s_addr = INADDR_ANY;
-  error_code ec = Bind((struct sockaddr*)&server_addr, sizeof(server_addr));
-  if (ec)
-    return ec;
-  return Listen(backlog);
+  if (LocalEndpoint().address().is_v4()) {
```
romange, line 98: have you tested it?
can you please create a unit test inside fiber_socket_test that tests these changes?
can also be inside accept_server_test - whatever is more convenient for you.
please note that our CI probably does not support ipv6 by default, but maybe it can be fixed - we did something in dragonfly repo to support. 

Minimal requirement: the unit test should cover ipv6 on a machine that has ipv6 enabled and skip the test on the machine that does not.

romange (commented):
lgtm, besides the test issue

util/fibers/fiber_socket_base.cc
vyavdoshenko: I tested this code manually. 
I will add tests.

romange (commented):
It's good that echo_server supports ipv6 now but why not add the test to unit tests?

vyavdoshenko:
> It's good that echo_server supports ipv6 now but why not add the test to unit tests?

I am in progress. I try to find all possible entities to test and add them.
I will notify you as soon as I finish adding everything.

util/fibers/fiber_socket_base.cc
vyavdoshenko: @romange 
I have updated the unit tests and tried adding IPv6 support everywhere.
Please take a look.

romange (approved):

romange:
lgtm but please fix compile error on macos:
https://github.com/romange/helio/actions/runs/13857796262/job/38788394190?pr=391#step:5:1649

vyavdoshenko:
> lgtm but please fix compile error on macos: https://github.com/romange/helio/actions/runs/13857796262/job/38788394190?pr=391#step:5:1649

Build has been fixed. I verified for macOS and Linux as well.

romange (approved):

PR #397, opened 2025-03-27
fix: short write in the tls socket

romange:
The bug is in TlsSocket::HandleUpstreamWrite and it went undetected
because it requires a mixture of interactions that are hard to achieve locally with regression tests.

This PR sinthesizes the same conditions of concurrent write and consume on TlsEngine that result in short write.

Signed-off-by: Roman Gershman <romange@gmail.com>

### DIFF
--- a/util/tls/tls_socket.cc
+++ b/util/tls/tls_socket.cc
@@ -18,6 +18,8 @@
                   << ", write_total:" << upstream_write_ << " "               \
                   << " pending output: " << engine_->OutputPending() << " "
 
+#define DVSOCK(verbosity) DVLOG(verbosity) << "sock[" << native_handle() << "] "
+
 namespace util {
 namespace tls {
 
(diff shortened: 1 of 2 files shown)


util/tls/tls_socket.cc
```diff
@@ -428,15 +430,16 @@ error_code TlsSocket::HandleUpstreamWrite() {
...
-  if (engine_->OutputPending() > 0) {
-    LOG(INFO) << "ssl buffer is not empty with " << engine_->OutputPending()
-              << " bytes. short write detected";
-  }
+    // We could preempt while calling WriteSome, and the engine could get more data to write.
+    // Therefore we sync the buffer.
+    buffer = engine_->PeekOutputBuf();
+  } while (!buffer.empty());
+
+  DCHECK_EQ(engine_->OutputPending(), 0u);
```
kostasrim, line 442: Should we also log this if the condition is not true just in case since `DCHECK` only works for debug ? On the other hand, we force the drain by Peeking above so :shrug:

romange: yeah, DCHECK here is more like a helpful comment

PR #398, opened 2025-04-06
chore: minor fixes

### LINKED ISSUE #1: statx is not implemented by musl-dev

romange:
1. Fix cmake warnings for newer cmake version (alpine)
2. Change StatFiles to use statx on linux. fixes #1


kostasrim (approved):
lgtm

Issue #399, opened 2025-04-06
provide a better scheduling policy for fibers than FIFO


romange:
## Use Case: Background Fibers

We sometimes need to execute a process in the background without impacting the priority of foreground fibers. Currently, we lack adequate tooling for this because the only available scheduling mechanism is FIFO. Simply yielding from a fiber is an insufficient solution, as that fiber can resume execution much faster than other fibers that are frequently blocked on I/O operations or mutexes.

## Solution

Implement two priority levels for fibers: regular and background.

* **Regular fibers:** These will operate similarly to the current implementation, maintaining a FIFO order among themselves.
* **Background fibers:** These will execute only after all regular fibers have finished running, effectively having secondary precedence. They will also maintain a FIFO order among themselves.

To prevent starvation of background fibers, the scheduler will track how long they have been bypassed by foreground fibers. If this duration exceeds a certain threshold, the scheduler will provide an opportunity for background fibers to run. This ensures that even under 100% CPU utilization, background fibers will still make progress. If CPU utilization is less than 100%, foreground fibers will eventually suspend, allowing background fibers to naturally take over execution.

PR #400, opened 2025-04-06
feat: introduce FiberPriority

romange:
Also, perform some clean-up in the code i touched. FiberPriority is not used yet, it is only stored in FiberInterface. The only functional change is that it fibers use boost fixedsize_stack, we now pass the stack size through the Opts struct, which is 64KB, while default boost fixedsize_stack size was 128KB, hence the change in fibers_test.cc.

First step for #399

### DIFF
--- a/util/fibers/detail/fiber_interface.cc
+++ b/util/fibers/detail/fiber_interface.cc
@@ -27,11 +27,13 @@ struct TL_FiberInitializer;
 
 namespace {
 
+[[maybe_unused]] size_t kSzFiberInterface = sizeof(FiberInterface);
+
 // Serves as a stub Fiber since it does not allocate any stack.
 // It's used as a main fiber of the thread.
 class MainFiberImpl final : public FiberInterface {
  public:
-  MainFiberImpl() noexcept : FiberInterface{MAIN, 1, "main"} {
+  MainFiberImpl() noexcept : FiberInterface{MAIN, FiberPriority::NORMAL, 1, "main"} {
   }
 
   ~MainFiberImpl() {
@@ -128,8 +130,8 @@ FiberInterface* FiberActive() noexcept {
   return FbInitializer().active;
 }
 
-FiberInterface::FiberInterface(Type type, uint32_t cnt, string_view nm)
-    : use_count_(cnt), type_(type) {
+FiberInterface::FiberInterface(Type type, FiberPriority prio, uint32_t cnt, string_view nm)
+    : use_count_(cnt), type_(type), prio_(prio) {
   remote_next_.store((FiberInterface*)kRemoteFree, memory_order_relaxed);
   size_t len = std::min(nm.size(), sizeof(name_) - 1);
   name_[len] = 0;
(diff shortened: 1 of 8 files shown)


util/fibers/fiber_socket_test.cc
```diff
@@ -209,19 +209,21 @@ TEST_P(FiberSocketTest, Basic) {
...
     std::ignore = sock->Connect(listen_ep_);
     accept_fb_.Join();
 
     // Explode the socket with Writes until it blocks. At this point the socket
     // should timeout;
+    constexpr size_t kBufSize = 1024 * 100;
+    std::unique_ptr<uint8_t[]> buf(new uint8_t[kBufSize]);
+    memset(buf.get(), 0, kBufSize);
     for (int i = 0; i < 50; ++i) {
-      uint8_t buf[1024 * 100] = {0};
```
romange, line 223: we should not create 100KB arrays on stack

util/fibers/fiber_socket_test.cc
```diff
@@ -209,19 +209,21 @@ TEST_P(FiberSocketTest, Basic) {
 }
 
 TEST_P(FiberSocketTest, Bug363) {
-  unique_ptr<FiberSocketBase> sock(proactor_->CreateSocket());
-  sock->set_timeout(3);
-
   proactor_->Await([&] {
     ThisFiber::SetName("ConnectFb");
+    unique_ptr<FiberSocketBase> sock(proactor_->CreateSocket());
```
romange, line 214: sockets should be created inside their threads

util/fibers/fibers_test.cc
```diff
@@ -215,7 +215,7 @@ TEST_F(FiberTest, Basic) {
   Fiber fb1("test1", [&] { ++run; });
   Fiber fb2("test2", [&] { ++run; });
   EXPECT_EQ(epoch, FiberSwitchEpoch());
-  EXPECT_EQ(128 * 1024 * 2, WorkerFibersStackSize());
+  EXPECT_EQ(64 * 1024 * 2, WorkerFibersStackSize());  // 2 fibers, 64k each
```
romange, line 218: the only change: default fiber stack is now 64KB

util/fibers/detail/fiber_interface.cc
```diff
@@ -27,11 +27,13 @@ struct TL_FiberInitializer;
 
 namespace {
 
+[[maybe_unused]] size_t kSzFiberInterface = sizeof(FiberInterface);
```
vyavdoshenko, line 30: Is this needed?

util/fibers/detail/fiber_interface.h
```diff
@@ -262,6 +269,7 @@ class FiberInterface {
   // A timestamp counter when this fiber became ready,active or got suspended (in cycles).
   uint64_t cpu_tsc_ = 0;
   char name_[24];
+  FiberPriority prio_;
```
vyavdoshenko, line 272: I suggest adding default initialization. NORMAL?

vyavdoshenko (approved):
lgtm

util/fibers/detail/fiber_interface.cc
romange: it's not needed, but it's a harmless hack that I use to display the struct size in vscode.

util/fibers/detail/fiber_interface.h
romange: it has a single c'tor that initializes this variable anyway, so defaul init here is redundant. I do not have a strong opinion but not worth another iteration, imho.

PR #401, opened 2025-04-06
chore: refactor scheduling code

romange:
1. Move scheduling step into Scheduler::RunWorkerFibersStep from Epoll/Uring proactors.
2. Extend ProactorBase::Await to support Fiber::Opts

### DIFF
--- a/util/fibers/detail/scheduler.cc
+++ b/util/fibers/detail/scheduler.cc
@@ -11,6 +11,7 @@
 
 #include "base/logging.h"
 #include "util/fibers/detail/utils.h"
+#include "util/fibers/proactor_base.h"
 #include "util/fibers/stacktrace.h"
 
 namespace util {
@@ -199,6 +200,12 @@ Scheduler::~Scheduler() {
   DestroyTerminated();
 }
 
+void Scheduler::Yield(FiberInterface* me) {
+  AddReady(me);
+  was_yield_ = true;
+  Preempt();
+}
+
 ctx::fiber_context Scheduler::Preempt() {
   if (FiberActive() == dispatch_cntx_.get()) {
     LOG(DFATAL) << "Should not preempt dispatcher: " << GetStacktrace();
@@ -210,7 +217,8 @@ ctx::fiber_context Scheduler::Preempt() {
     if (now != last_ts) {  //   once a second at most.
       last_ts = now;
       LOG(DFATAL) << "Preempting inside of atomic section, fiber: " << FiberActive()->name()
-                << ", stacktrace:\n" << GetStacktrace();
+                  << ", stacktrace:\n"
+                  << GetStacktrace();
     }
   }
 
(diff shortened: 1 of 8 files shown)


util/fibers/detail/scheduler.cc
```diff
@@ -484,6 +492,24 @@ void Scheduler::PrintAllFiberStackTraces() {
   ExecuteOnAllFiberStacks(print_fn);
 }
 
+bool Scheduler::RunWorkerFibersStepImpl() {
+  DCHECK(!ready_queue_.empty());
+
+  FiberInterface* fi = PopReady();
```
romange, line 498: the code is moved from xxx_proactor.cc as is

util/fibers/detail/scheduler.h
```diff
@@ -132,12 +148,13 @@ class Scheduler {
   FI_Queue ready_queue_, terminate_queue_;
   SleepQueue sleep_queue_;
   base::MPSCIntrusiveQueue<FiberInterface> remote_ready_queue_;
-  std::atomic_uint64_t remote_epoch_{0};
```
romange, line 135: was not used

util/fibers/proactor_base.h
```diff
@@ -104,6 +104,11 @@ class ProactorBase {
     return tl_info_.monotonic_time;
   }
 
+  // Used by Scheduler to update the monotonic time.
+  static void UpdateMonotonicTime() {
+    tl_info_.monotonic_time = GetClockNanos();
```
vyavdoshenko, line 109: I am just curious. Should monotonic_time be atomic?
It seems it should because it can be read from different threads.

vyavdoshenko (approved):
lgtm

util/fibers/proactor_base.h
kostasrim: `__thread ProactorBase::TLInfo ProactorBase::tl_info_;`

romange: tl_info_ is thread local so it can not be accessed from a foreign thread.

util/fibers/detail/scheduler.cc
```diff
@@ -484,6 +492,24 @@ void Scheduler::PrintAllFiberStackTraces() {
...
+  DCHECK(!fi->list_hook.is_linked());
+  DCHECK(!fi->sleep_hook.is_linked());
+  AddReady(dispatch_cntx_.get());
+
+  DVLOG(2) << "Switching to " << fi->name();
+  ProactorBase::UpdateMonotonicTime();
+  fi->SwitchTo();
+
+  was_yield_ = false;
```
kostasrim, line 507: Should be true right ? We yieled, we switched to a different fiber

kostasrim (commented):
minor comment lgtm

util/fibers/detail/scheduler.cc
kostasrim: Oh nvm, we set it to true when we actually yield. Got it.

romange: no, `was_yield_` is set to true  by a fiber that yields. here we reset it back. will clarify in the next PRs

PR #402, opened 2025-04-08
feat(scheduler): support scheduling of background fibers

### LINKED ISSUE #399: provide a better scheduling policy for fibers than FIFO

romange:
Fixes #399

### DIFF
--- a/util/fibers/detail/scheduler.cc
+++ b/util/fibers/detail/scheduler.cc
@@ -6,6 +6,7 @@
 #include <absl/strings/str_cat.h>
 #include <absl/time/clock.h>
 
+#include <chrono>
 #include <condition_variable>
 #include <mutex>
 
@@ -22,6 +23,9 @@ namespace ctx = boost::context;
 
 using namespace std;
 using base::CycleClock;
+using chrono::duration;
+using chrono::steady_clock;
+
 namespace {
 
 class DispatcherImpl final : public FiberInterface {
@@ -123,8 +127,8 @@ void DispatcherImpl::DefaultDispatch(Scheduler* sched) {
       sched->ProcessSleep();
     }
 
-    if (sched->HasReady()) {
-      FiberInterface* fi = sched->PopReady();
+    if (sched->HasReady(this->prio_)) {
+      FiberInterface* fi = sched->PopReady(this->prio_);
       DCHECK(!fi->list_hook.is_linked());
       DCHECK(!fi->sleep_hook.is_linked());
       sched->AddReady(this);
@@ -169,11 +173,12 @@ Scheduler::~Scheduler() {
   shutdown_ = true;
   DCHECK(main_cntx_ == FiberActive());
 
-  const unsigned prio = unsigned(FiberPriority::NORMAL);
-  while (HasReady(prio)) {
-    FiberInterface* fi = PopReady(prio);
-    DCHECK(!fi->sleep_hook.is_linked());
-    fi->SwitchTo();
+  for (auto prio : {FiberPriority::NORMAL, FiberPriority::BACKGROUND}) {
+    while (HasReady(prio)) {
+      FiberInterface* fi = PopReady(prio);
+      DCHECK(!fi->sleep_hook.is_linked());
+      fi->SwitchTo();
+    }
   }
 
   DispatcherImpl* dimpl = static_cast<DispatcherImpl*>(dispatch_cntx_.get());
@@ -199,13 +204,10 @@ Scheduler::~Scheduler() {
   DestroyTerminated();
 }
 
-void Scheduler::Yield(FiberInterface* me) {
-  AddReady(me);
-  yield_occurred_ = true;
-  Preempt();
-}
+ctx::fiber_context Scheduler::Preempt(bool yield) {
+  // One in X times put background fibers to sleep to increase responsiveness
+  const size_t kBackgroundSleepFrequency = 10;
 
-ctx::fiber_context Scheduler::Preempt() {
   if (FiberActive() == dispatch_cntx_.get()) {
     LOG(DFATAL) << "Should not preempt dispatcher: " << GetStacktrace();
   }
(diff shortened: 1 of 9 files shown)


util/fibers/detail/scheduler.cc
```diff
@@ -521,24 +558,22 @@ auto Scheduler::RunWorkerFibersStepImpl() -> RunFiberResult {
     fi->SwitchTo();
 
     ProactorBase::UpdateMonotonicTime();
-    uint64_t now_ns = ProactorBase::GetMonotonicTimeNs();
+    ran_ns = ProactorBase::GetMonotonicTimeNs() - start_ns;
 
     // It could be that we iterate over multiple fibers and meanwhile we postpone the execution of
     // brief tasks or do not handle I/O events. Therefore we limit the time spent in the loop
     // to kMaxTimeSpentNs.
-    if (now_ns > deadline_ns) {
```
romange, line 529: why you did not like `deadline_ns` ? it saved me a substract for each iteration :)

util/fibers/detail/scheduler.cc
```diff
@@ -521,24 +555,22 @@ auto Scheduler::RunWorkerFibersStepImpl() -> RunFiberResult {
     fi->SwitchTo();
 
     ProactorBase::UpdateMonotonicTime();
-    uint64_t now_ns = ProactorBase::GetMonotonicTimeNs();
+    ran_ns = ProactorBase::GetMonotonicTimeNs() - start_ns;
```
romange, line 558: so since we always preempt into NORMAL queue, this measurement is incorrect. imagine fiber F1 - background, fiber F2..F10 - are normal. F10 is being the dispatch fiiber. F1 runs, then switches to F2 and so on until it returns to dispatcher. Everything will be accounted for as `ran_ns` here.

romange: you should decide what you do when a BACKGROUND fiber preempts. Do we continiue with normal fibers if they exist? Do we switch back to the dispatcher fiber? Continue with other background fibers?

util/fibers/detail/scheduler.cc
```diff
@@ -508,10 +545,13 @@ auto Scheduler::RunWorkerFibersStepImpl() -> RunFiberResult {
   // To mitigate this,
   // we break out of the loop if a fiber yield has occurred, even if the ready queue is not empty.
   do {
-    FiberInterface* fi = PopReady(q_indx);
+    FiberInterface* fi = PopReady(priority);
     DCHECK(!fi->list_hook.is_linked());
     DCHECK(!fi->sleep_hook.is_linked());
-    AddReady(dispatch_cntx_.get());
+
+    // Run round robin only with equal priority, others switch back directly after running
```
romange, line 552:
```suggestion
    // dispatch fiber is added only to the normal queue, for background queue we return back to dispatch 
    // after running a single background fiber
```

romange:
requires rebase

util/fibers/detail/scheduler.h
```diff
@@ -151,16 +162,27 @@ class Scheduler {
 
   boost::intrusive_ptr<FiberInterface> dispatch_cntx_;
 
-  // ready_queue_[0] - normal. TODO: add background queue.
-  FI_Queue ready_queue_[1], terminate_queue_;
+
+  struct  {
+    uint64_t last_normal_run = 0;
```
romange, line 167: can you please add comments explaining each field?

util/fibers/uring_proactor.cc
```diff
@@ -931,6 +941,7 @@ void UringProactor::MainLoop(detail::Scheduler* scheduler) {
     bool should_poll =
         spin_loops++ < 10 || (base::CycleClock::Now() < idle_end_cycle() + busy_poll_cycles);
     if (!ring_busy && should_poll) {
+      // CHECK(false);
```
romange, line 944: remove

util/fibers/detail/scheduler.cc
```diff
@@ -200,8 +205,7 @@ Scheduler::~Scheduler() {
 }
 
 void Scheduler::Yield(FiberInterface* me) {
-  AddReady(me);
-  yield_occurred_ = true;
+  round_robin_run_.last_yield = true;
```
romange, line 208: Seems that it's simpler to introduce `Preempt(bool should_yield)` instead of having this state variable `last_yield`

util/fibers/detail/scheduler.cc
```diff
@@ -339,6 +378,7 @@ bool Scheduler::WaitUntil(chrono::steady_clock::time_point tp, FiberInterface* m
   return has_timed_out;
 }
 
+// TODO: Optimize
```
romange, line 381: not clear the intent. either expand or remove

romange: ?

util/fibers/detail/scheduler.cc
```diff
@@ -494,53 +534,89 @@ void Scheduler::PrintAllFiberStackTraces() {
   ExecuteOnAllFiberStacks(print_fn);
 }
 
-auto Scheduler::RunWorkerFibersStepImpl() -> RunFiberResult {
-  const unsigned q_indx = unsigned(FiberPriority::NORMAL);
-  DCHECK(HasReady(q_indx));
+RunFiberResult Scheduler::Run(FiberPriority priority) {
```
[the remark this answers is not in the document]
romange: was it handled @dranikpg ?

util/fibers/detail/scheduler.cc
dranikpg: Oh, yes, sorry

dranikpg: There's a 100 iteration loop that I'd like to get rid of 🤓

dranikpg (approved):

dranikpg (approved):

PR #404, opened 2025-04-10
chore: CycleClock clean ups

romange:
1. Remove redundant CycleClock code that I have not cleaned up before
2. Expose `ThisFiber::GetRunningTimeCycles` to be able to check for how long the current fiber is running.


base/init.cc
```diff
@@ -60,7 +60,14 @@ namespace base {
 uint64_t CycleClock::frequency_ = 0;
 
 void CycleClock::InitOnce() {
+#ifdef __aarch64__
```
vyavdoshenko, line 63: I am just curious.
rdtsc for Intel work not well?

vyavdoshenko (approved):
lgtm

vyavdoshenko (approved):

PR #407, opened 2025-05-07
feat: Add TCP socket info retrieval to proc_reader module

vyavdoshenko:
This PR adds new functionality to the proc_reader module for retrieving TCP socket information from /proc/net/tcp and /proc/net/tcp6 files.
The functionality enables retrieving detailed information about TCP sockets by their inode numbers, which is useful for diagnostics and monitoring.

### DIFF
--- a/io/proc_reader.cc
+++ b/io/proc_reader.cc
@@ -87,6 +87,164 @@ size_t find_nth(string_view str, char c, uint32_t index) {
   return string_view::npos;
 }
 
+// Convert hex string to IPv6 address bytes
+void HexToIPv6(string_view hex_str, unsigned char* out) {
+  for (size_t i = 0; i < 16 && (i * 2 + 1) < hex_str.size(); i++) {
+    string_view byte_hex = hex_str.substr(i * 2, 2);
+    unsigned int byte;
+    if (absl::SimpleHexAtoi(byte_hex, &byte)) {
+      out[i] = static_cast<unsigned char>(byte);
+    }
+  }
+}
+
+// Parse a line from /proc/net/tcp or /proc/net/tcp6 file
+bool ParseSocketLine(string_view line, ino_t target_inode, bool is_ipv6, TcpInfo* info) {
+  line = absl::StripLeadingAsciiWhitespace(line);
+
+  std::vector<string_view> parts;
+  size_t pos = 0;
+
+  while (pos < line.size()) {
+    while (pos < line.size() && absl::ascii_isspace(line[pos]))
+      pos++;
+    if (pos >= line.size())
+      break;
+
+    size_t token_start = pos;
+    while (pos < line.size() && !absl::ascii_isspace(line[pos]))
+      pos++;
+
+    parts.push_back(line.substr(token_start, pos - token_start));
+  }
+
+  if (parts.size() < 10)
+    return false;
+
+  unsigned int sl;
+  string_view sl_part = parts[0];
+  if (absl::EndsWith(sl_part, ":")) {
+    sl_part.remove_suffix(1);
+  }
+  if (!absl::SimpleAtoi(sl_part, &sl))
+    return false;
+
+  string_view local_addr_port = parts[1];
+  size_t colon_pos = local_addr_port.find(':');
+  if (colon_pos == string_view::npos)
+    return false;
+
+  string_view local_addr_hex = local_addr_port.substr(0, colon_pos);
+  string_view local_port_hex = local_addr_port.substr(colon_pos + 1);
+
+  string_view remote_addr_port = parts[2];
+  colon_pos = remote_addr_port.find(':');
+  if (colon_pos == string_view::npos)
+    return false;
+
+  string_view remote_addr_hex = remote_addr_port.substr(0, colon_pos);
+  string_view remote_port_hex = remote_addr_port.substr(colon_pos + 1);
+
+  unsigned int state;
+  if (!absl::SimpleHexAtoi(parts[3], &state))
+    return false;
+
+  unsigned int inode = 0;
+  if (!absl::SimpleAtoi(parts[9], &inode))
+    return false;
+
+  if (inode != target_inode)
+    return false;
+
+  info->is_ipv6 = is_ipv6;
+  info->state = state;
+  info->inode = inode;
+
+  unsigned int port;
+  if (absl::SimpleHexAtoi(local_port_hex, &port)) {
+    info->local_port = port;
+  }
+  if (absl::SimpleHexAtoi(remote_port_hex, &port)) {
+    info->remote_port = port;
+  }
+
+  if (is_ipv6) {
+    HexToIPv6(local_addr_hex, info->local_addr6);
+    HexToIPv6(remote_addr_hex, info->remote_addr6);
+  } else {
+    unsigned int addr;
+    if (absl::SimpleHexAtoi(local_addr_hex, &addr)) {
+      info->local_addr = addr;
+    }
+    if (absl::SimpleHexAtoi(remote_addr_hex, &addr)) {
+      info->remote_addr = addr;
+    }
+  }
+
+  return true;
+}
+
+// Reads TCP info from a specified proc file
+Result<TcpInfo> ReadTcpInfoFromFile(const char* proc_path, ino_t sock_inode, bool is_ipv6) {
+  TcpInfo info;
+  int fd = open(proc_path, O_RDONLY | O_CLOEXEC);
+  if (fd < 0) {
+    return make_unexpected(error_code{errno, system_category()});
+  }
+
+  base::IoBuf buf{4096};
+  bool found = false;
+  bool header_skipped = false;
+
+  while (true) {
+    auto dest = buf.AppendBuffer();
+
+    int res = read(fd, dest.data(), dest.size());
+    if (res < 0) {
+      close(fd);
+      return make_unexpected(error_code{errno, system_category()});
+    }
+
+    if (res == 0) {
+      break;
+    }
+
+    buf.CommitWrite(res);
+
+    while (true) {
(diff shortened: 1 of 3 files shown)


io/proc_reader.cc
```diff
@@ -87,6 +87,166 @@ size_t find_nth(string_view str, char c, uint32_t index) {
...
+  }
+
+  base::IoBuf buf{4096};
+  bool found = false;
+  bool header_skipped = false;
+
+  while (true) {
+    auto dest = buf.AppendBuffer();
+
+    int res = read(fd, dest.data(), dest.size());
```
kostasrim, line 204: Helio has utilities to read and write to a file -- you can use these and work well with the async interface of iouring. As is, we have a `blocking` call, meaning that it blocks the thread until the operation is done. Furthermore, the open without the O_DIRECT goes through the page cache. What we get here is buffered IO.

Not sure if we are fine with these blocking calls. Maybe @romange can add some comment here

kostasrim: plz see `io/file.h` for how we do it inhelio

io/proc_reader.h
```diff
@@ -48,4 +49,20 @@ using DistributionInfo = std::vector<std::pair<std::string, std::string>>;
...
+  unsigned remote_port = 0;
+  unsigned inode = 0;
+  uint32_t local_addr = 0;
+  uint32_t remote_addr = 0;
+  unsigned char local_addr6[16] = {0};
+  unsigned char remote_addr6[16] = {0};
+};
+
+std::string TcpStateToString(unsigned state);
+Result<TcpInfo> ReadTcpInfo(ino_t sock_inode);
```
romange, line 65: add a comment here that sock_inode can be fetched with fstat

io/proc_reader.cc
```diff
@@ -87,6 +87,166 @@ size_t find_nth(string_view str, char c, uint32_t index) {
...
+    }
+  }
+
+  return true;
+}
+
+// Reads TCP info from a specified proc file
+Result<TcpInfo> ReadTcpInfoFromFile(const char* proc_path, ino_t sock_inode, bool is_ipv6) {
+  TcpInfo info;
+  int fd = open(proc_path, O_RDONLY | O_CLOEXEC);
```
romange, line 192: you should not reimplement it. you can just call `ReadProcFile('/proc/net/tcp', ':', your_parse_line_cb);`

the entire logic should be inside `your_parse_line_cb`

io/proc_reader.cc
romange: it's not needed here, as `/proc/` filesystem is special.

io/proc_reader.h
```diff
@@ -48,4 +49,20 @@ using DistributionInfo = std::vector<std::pair<std::string, std::string>>;
 
 Result<DistributionInfo> ReadDistributionInfo();
 
+struct TcpInfo {
+  bool is_ipv6 = false;
+  unsigned state = 0;
```
romange, line 54: add a comment with a url to page explaining the state values.

vyavdoshenko: fixed

io/proc_reader.h
vyavdoshenko: fixed

io/proc_reader.cc
vyavdoshenko: ReadProcFile reads each line like key/value using the separator, i.e., ':'.
/proc/net/tcp or /proc/net/tcp6 has a table structure. It means I can't reuse the function; it should be deeply updated to support not only pairs of values but also tuples of values.

@romange, do you want me to rewrite this function?

romange: it's not accurate. indeed it does not parse tuples but it still splits to lines and removes the first column.
so say my tcp file has:
```
sl  local_address rem_address   st tx_queue rx_queue tr tm->when retrnsmt   uid  timeout inode                                                     
   0: 0100007F:0277 00000000:0000 0A 00000000:00000000 00:00000000 00000000     0        0 6032246 1 0000000000000000 100 0 0 10 0                   
   1: 0100007F:B4AD 00000000:0000 0A 00000000:00000000 00:00000000 00000000     0        0 18723 1 0000000000000000 100 0 0 10 0                     
   2: 3600007F:0035 00000000:0000 0A 00000000:00000000 00:00000000 00000000   991        0 3764 1 0000000000000000 100 0 0 10 5                      

```

then with ':' separator,  the second argument  in the callback will get 
`0100007F:0277 00000000:0000 0A 00000000:00000000 00:00000000 00000000     0        0 6032246 1`

and yes, you will need to parse it but it saves you from duplicating the code around readin files and buffer management.

romange: ah , I see, it does not know how to skip headers.
you can add an argument `book skip_header` to ReadProcFile

vyavdoshenko: fixed

io/proc_reader.cc
```diff
@@ -87,6 +95,122 @@ size_t find_nth(string_view str, char c, uint32_t index) {
...
+
+  return true;
+}
+
+// Reads TCP info from a specified proc file
+Result<TcpInfo> ReadTcpInfoFromFile(const char* proc_path, ino_t sock_inode, bool is_ipv6) {
+  TcpInfo info;
+  bool found = false;
+
+  auto cb = [&](string_view key, string_view value) {
```
romange, line 192: nit, instead of referencing `found`, you could write:

`auto cb = [found = false](string_view key, string_view value) mutable {`

romange (approved):

io/proc_reader.cc
```diff
@@ -87,6 +95,122 @@ size_t find_nth(string_view str, char c, uint32_t index) {
...
+    }
+  }
+}
+
+// Parse a line from /proc/net/tcp or /proc/net/tcp6 file
+bool ParseSocketLine(string_view line, ino_t target_inode, bool is_ipv6, TcpInfo* info) {
+  std::vector<string_view> parts;
+  size_t pos = 0;
+
+  while (pos < line.size()) {
```
romange, line 114: you can replace this loop with `absl::StrSplit`

romange: you can also use absl::SkipEmpty to skip consecutive whitespace characters.

io/proc_reader.cc
```diff
@@ -87,6 +95,122 @@ size_t find_nth(string_view str, char c, uint32_t index) {
...
+    if (pos >= line.size())
+      break;
+
+    size_t token_start = pos;
+    while (pos < line.size() && !absl::ascii_isspace(line[pos]))
+      pos++;
+
+    parts.push_back(line.substr(token_start, pos - token_start));
+  }
+
```
romange, line 126: add a comment with the header describing all the parts (aka `local_address rem_address   st tx_queue rx_queue tr tm->when retrnsmt   uid  timeout inode `)

io/proc_reader.cc
```diff
@@ -87,6 +95,122 @@ size_t find_nth(string_view str, char c, uint32_t index) {
...
+  if (is_ipv6) {
+    HexToIPv6(local_addr_hex, info->local_addr6);
+    HexToIPv6(remote_addr_hex, info->remote_addr6);
+  } else {
+    unsigned int addr;
+    if (absl::SimpleHexAtoi(local_addr_hex, &addr)) {
+      uint8_t* bytes = reinterpret_cast<uint8_t*>(&addr);
+      info->local_addr = (bytes[0] << 24) | (bytes[1] << 16) | (bytes[2] << 8) | bytes[3];
```
romange, line 176: you can just use `ntohl()` or `htonl()` which is clearer.

romange: actually it's not related to it but there is `absl::big_endian::Load32`

io/proc_reader.cc
vyavdoshenko: I found the description: https://www.kernel.org/doc/Documentation/networking/proc_net_tcp.txt
It looks like it's reasonable to replace the implementation with tcp4_seq_show() and tcp6_seq_show().

romange: not sure I follow. it's kernel internal functions

io/proc_reader.cc
vyavdoshenko: fixed

io/proc_reader.cc
vyavdoshenko: fixed

io/proc_reader.cc
vyavdoshenko: fixed

io/proc_reader.cc
vyavdoshenko: The `found` is not the only variable that should be passed. There are 4 of them. I think the reference is more convenient here.

romange (approved):

io/proc_reader.cc
```diff
@@ -87,6 +97,108 @@ size_t find_nth(string_view str, char c, uint32_t index) {
...
+    }
+  }
+
+  return true;
+}
+
+// Reads TCP info from a specified proc file
+Result<TcpInfo> ReadTcpInfoFromFile(const char* proc_path, ino_t sock_inode, bool is_ipv6) {
+  TcpInfo info;
+  bool found = false;
```
romange, line 178:
```suggestion
```

io/proc_reader.cc
```diff
@@ -87,6 +97,108 @@ size_t find_nth(string_view str, char c, uint32_t index) {
...
+
+  auto cb = [&](string_view key, string_view value) mutable {
```
romange, line 180:
```suggestion
  auto cb = [found=false](string_view key, string_view value) mutable {
```

vyavdoshenko: I've already merged the PR.
Would you like me to polish this part of the code in a separate PR?

PR #410, opened 2025-05-09
chore: refactor code in accept_server

romange:
Moved lambda to be a free function. Also moved listener breaking code into ListenerInterface class for better encapsulation.

No functional changes.


vyavdoshenko (approved):
lgtm

PR #411, opened 2025-05-12
chore: refactor azure/gcp code

romange:
Small refactoring around cloud::CredentialsProvider so now Azure can rely on an abstract interface instead of using the concrete one.


vyavdoshenko (approved):
lgtm

PR #417, opened 2025-06-05
fix: Check if connection is closed and retry sending

mkaruza:


romange:
Have you checked it fixes the issue?

mkaruza:
Update PR so it handles `broken_pipe` which can happen after `req->Send()`

util/http/http_client.cc
```diff
@@ -133,6 +133,10 @@ bool Client::IsConnected() const {
   return socket_ && socket_->IsOpen();
 }
 
+bool Client::IsClosed(const std::error_code& ec) const {
+  return socket_ && socket_->IsConnClosed(ec);
```
romange, line 137: `FiberSocketBase::IsConnClosed` is a static function and you do not need this wrapper to call it.
just use it directly where you need.

romange (approved):

PR #418, opened 2025-06-06
fix: zero bytes sent error

BorysTheDev:
problem: if user sends 0 bytes, error is returned
fix: if 0 bytes is sent, check that buffers were empty

### DIFF
--- a/.github/workflows/mac-os.yml
+++ b/.github/workflows/mac-os.yml
@@ -45,4 +45,4 @@ jobs:
           ninja -k 5 base/all io/all strings/all util/all echo_server ping_iouring_server \
             https_client_cli s3_demo
           ./fibers_test --logtostderr --gtest_repeat=10
-          ctest -V -L CI
+          GLOG_logtostderr=1 ctest -V -L CI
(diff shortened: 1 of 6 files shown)


vyavdoshenko (approved):
lgtm

is it possible to add unit test for that?

BorysTheDev:
> I disagree with the approach. I would say that sending a zero length iovec is a bug and there is a bug in the caller. Maybe I am missing something, so I would love to hear reasons why would a caller send zero bytes

Imagine you have a format where you should write field sizes first and then the data of these fields (tungsten format for example) so when you want to send these format you will have the following iovec {&field1_size, sizeof(uint64)}, {&field2_size, sizeof(uint64)}, ..., {&field1_data, field1_size}, {&field2_data, field2_size}... I don't see any problem if some data is 0 size. And sending 0 bytes looks correct

romange:
It should not be sent to kernel and we should not handle though. In many
cases we filter out this in the higher levels. So either here before the
call or at the caller level or both.

Best regards,
     Roman

romange:
I still prefer to skip an empty iovec, it's a single condition, and pretty straightforward . Moreover, I wish we will add a dcheck to catch the caller who sends it

romange:
Ok. Please also fix logtraffic

PR #422, opened 2025-06-09
chore(io): add a sanity check that iovec is not empty

romange:


io/io.cc
```diff
@@ -16,7 +16,9 @@ namespace io {
 namespace {
 
 // Writes some bytes and handles some corner cases including retry on interrupted.
-inline Result<size_t> WriteSomeBytes(const iovec* v, uint32_t len, Sink* dest) {
+Result<size_t> WriteSomeBytes(const iovec* v, uint32_t len, Sink* dest) {
+  DCHECK(len > 0 && v != nullptr);
+  DCHECK_GT(v->iov_len, 0u);
```
BorysTheDev, line 21: It doesn't work, because iovec is array with size len, so if you want such DCHECK you need
for (int i = 0; i < len; ++i)
  DCHECK_GT(v[i].size, 0u);

BorysTheDev: it is quite confusing because  v[o].iov_len == 0 and v[1].iov_len != 0 we notify dcheck but if v[o].iov_len != 0 and v[1].iov_len == 0 it's ok

romange: fixed

PR #424, opened 2025-06-17
 chore: write/read in progress for tls socket

kostasrim:


kostasrim:
follow up from https://github.com/romange/helio/pull/416 

one more PR to clean everything up and tls async functionality should be complete

p.s. need to rebase, as is it includes/shows the diff including the changes in pull 416

PR #431, opened 2025-07-12
fix: SSL_shutdown on incomplete handshake

### LINKED ISSUE #430: TlsSocket shutdown while in init
`SSL protocol error 107:336462231 shutdown while in init ../helio/util/tls/tls_engine.cc : 150 `

The problem is that we call `shutdown` during the `handshake` phase.

kostasrim:
We should not call `SSL_shutdown` if the handshake is not complete.

Furthermore reduce the `LOG(INFO)` because they can flood and steal cpu time.

Fixes: #430

### DIFF
--- a/util/tls/tls_engine.cc
+++ b/util/tls/tls_engine.cc
@@ -143,7 +143,7 @@ auto Engine::Shutdown() -> OpResult {
   int result = SSL_shutdown(ssl_);
   // See https://www.openssl.org/docs/man1.1.1/man3/SSL_shutdown.html
 
-  if (result == 0) { // First step of Shutdown (close_notify) returns 0.
+  if (result == 0) {              // First step of Shutdown (close_notify) returns 0.
     result = SSL_shutdown(ssl_);  // Initiate the second step.
   }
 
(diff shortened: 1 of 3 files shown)


util/tls/tls_engine.cc
```diff
@@ -237,15 +236,16 @@ auto Engine::ToOpResult(int result, const char* location) -> OpResult {
...
       queue_error = ERR_get_error();
-      LOG(WARNING) << "SSL protocol error " << ERROR_DETAILS;
+      LOG_EVERY_T(WARNING, 30) << "SSL protocol error " << ERROR_DETAILS;
       break;
+    }
     default:
       queue_error = ERR_get_error();
       state_ |= FATAL_ERROR;
-      LOG(WARNING) << "Unexpected SSL error " << ssl_error << " " << ERROR_DETAILS;
+      LOG_EVERY_T(WARNING, 30) << "Unexpected SSL error " << ssl_error << " " << ERROR_DETAILS;
```
romange, line 248: same

romange (approved):

PR #447, opened 2025-08-05
chore: busy polling specific per proactor

romange:
Beforehand we had a flag that was common to all proactors. Now the responsibility to override goes to caller and each proactor can have its own setting.

### DIFF
--- a/util/fibers/proactor_base.cc
+++ b/util/fibers/proactor_base.cc
@@ -24,7 +24,7 @@ constexpr int kNumSig = _NSIG;
 
 #else
 constexpr int kNumSig = NSIG;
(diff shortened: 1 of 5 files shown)


util/fibers/proactor_base.cc
```diff
@@ -73,11 +73,9 @@ static inline void arm_arch_pause(void) {
 
   if (use_spin_delay_sb == 1) {
     asm volatile(".inst 0xd50330ff");  // SB instruction encoding
-  }
```
romange, line 76: whitespace

BorysTheDev (approved):

PR #449, opened 2025-08-11
fix(scheduler): HasReady checks after different priority wakeups

dranikpg:


util/fibers/detail/scheduler.cc
```diff
@@ -546,18 +546,23 @@ RunFiberResult Scheduler::Run(FiberPriority priority) {
     }
   }
 
-  auto res = HasReady(priority) ? RunReadyFibersInternal(priority) : RunFiberResult::EXHAUSTED;
+  // Proactor should progress on normal priority fibers first
+  DCHECK(priority != FiberPriority::BACKGROUND || !HasReady(FiberPriority::NORMAL));
```
romange, line 550: consider `DCHECK(priority == FiberPriority::NORMAL || !HasReady(FiberPriority::NORMAL));`

romange (approved):

PR #452, opened 2025-08-14
Change background_sleep_freq to be %

dranikpg:
And yield only if not idle


util/fibers/detail/scheduler.cc
```diff
@@ -235,11 +235,13 @@ ctx::fiber_context Scheduler::Preempt(bool yield) {
     bool hungry = round_robin_run_.last_ts - last > round_robin_run_.budget_ns;
     // No other normal fibers were present in last 5ms, server is likely (almost) idle
     bool likely_idle = round_robin_run_.last_ts - round_robin_run_.last_normal_ts > 5'000'000;
+    // Yield with specified probability (frequency)
+    bool freq_yield = round_robin_run_.last_ts % 100 < config_.background_sleep_freq;
```
romange, line 239: are you sure you do not want to do any mixing/hashing on `round_robin_run_.last_ts` ? it could have some bias that would have catastrophic result on the algo. Also, I find `frequency` name confusing because it does look like probability.

romange (dismissed):

PR #455, opened 2025-08-20
fix: separate msgring support and enable states

romange:
This is needed to prevent the situation when we enable msgring on platforms that do not support this API. See https://github.com/dragonflydb/dragonfly/issues/5693 for more context.


romange:
yes, it should fix it once we update helio

PR #457, opened 2025-08-25
chore: add high priority to the scheduler

romange:
High priority share the NORMAL queue with NORMAL fibers.

If a high priority fiber is activated, it is pushed to the front of the queue. If it yields, however, it is still pushed to back.

Also added support for priorities for default, main thread that does not run an I/O polling loop. It is mainly added for being able to test fibers without proactors.


vyavdoshenko (approved):
IMHO, a separate high-priority queue could be better.
Interesting, can NORMAL priority start starving?

lgtm

PR #461, opened 2025-09-03
chore:include DISPATCH fiber in long running warnings

romange:
1. Fix the issue that for many cases we reset cpu_tsc_ incorrectly inside AddReady function. cpu_tsc_ measures the state start for differrent states. It could be for when we add a fiber into a ready queue, and sometimes when the fiber starts running. This allows us to measure the time once the state transitions. Before we reset cpu_tsc_ inside AddReady so during yield operations we reset cpu_tsc_ while we are running, disrupting the RUNNING state measurement when we suspend. Now we reset cpu_tsc_ outside of AddReady and only for non-yield scenarios.

2. Also improve logging precision of long running fibers.


vyavdoshenko (approved):
lgtm

PR #465, opened 2025-09-13
fix: mac-os build failures

romange:
Also, handle gracious close of http connection on a storage server if it asks for it (reuse_connecton is false).


dranikpg (approved):
👍🏻 I upgraded cmake locally as well and now define EXPORT_....=3.5 as an env globally

PR #469, opened 2025-09-22
chore: add fsync and statx in SubmitEntry

kostasrim:
* add fsync in SubmitEntry
* add statx in SubmitEntry
* fix fallocate arguments in SubmitEntry
* add blocking fsync and statx to UringFiler

### DIFF
--- a/util/fibers/submit_entry.h
+++ b/util/fibers/submit_entry.h
@@ -130,6 +130,22 @@ class SubmitEntry {
     sqe_->off = offset;
   }
 
+  void PrepFSync(int fd, unsigned flags) {
+    PrepFd(IORING_OP_FSYNC, fd);
+    sqe_->fsync_flags = flags;
+  }
+
+  void PrepStatX(const char* filepath, struct statx *stat) {
+    // AT_FDCWD is ignored when addr is an absolute path
+	  PrepFd(IORING_OP_STATX, AT_FDCWD);
+    sqe_->off = reinterpret_cast<uint64_t>(stat);
+    sqe_->addr = reinterpret_cast<unsigned long>(filepath);
+    // mask
+    sqe_->len = STATX_BASIC_STATS;
+    sqe_->statx_flags = 0;
+  }
+
+
   void PrepSend(int fd, const void* buf, size_t len, unsigned flags) {
     PrepFd(IORING_OP_SEND, fd);
     sqe_->addr = (__u64)buf;
--- a/util/fibers/uring_file.h
+++ b/util/fibers/uring_file.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <sys/types.h>  // for mode_t
+#include <sys/stat.h> // statx
 
 #include "io/file.h"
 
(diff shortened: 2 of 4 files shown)


util/fibers/uring_file.h
```diff
@@ -75,6 +79,11 @@ class LinuxFile {
   void WriteFixedAsync(io::Bytes src, off_t offset, unsigned buf_index, AsyncCb cb);
   void WriteAsync(io::Bytes src, off_t offset, AsyncCb cb);
   void FallocateAsync(int mode, off_t offset, off_t len, AsyncCb cb);
+
+#ifndef __APPLE__
+  std::error_code StatX(std::string_view filepath, struct statx *stat);
```
kostasrim, line 84: This interface is kinda bad, because it accepts the `filepath` which should be stored in `UringFile` but we don't 🤷‍♂️

util/fibers/submit_entry.h
```diff
@@ -119,8 +120,8 @@ class SubmitEntry {
   void PrepFallocate(int fd, int mode, off_t offset, off_t len) {
     PrepFd(IORING_OP_FALLOCATE, fd);
     sqe_->off = offset;
-    sqe_->len = mode;
-    sqe_->addr = uint64_t(len);
+    sqe_->len = len;
```
kostasrim, line 123: How did this work 🤷‍♂️

romange: it worked . this is the implementation of liburing library:
```
IOURINGINLINE void io_uring_prep_fallocate(struct io_uring_sqe *sqe, int fd,
					   int mode, __u64 offset, __u64 len)
{
	io_uring_prep_rw(IORING_OP_FALLOCATE, sqe, fd,
			0, (unsigned int) mode, (__u64) offset);
	sqe->addr = (__u64) len;
}

```
so I do not think yours is correct. In any case, please add coverage to util/fibers/uring_file_test.cc

util/fibers/submit_entry.h
```diff
@@ -130,6 +131,22 @@ class SubmitEntry {
...
 
+  void PrepFSync(int fd) {
+    PrepFd(IORING_OP_FSYNC, fd);
+    // TODO fsync with flags = fdatasync
+    sqe_->sync_range_flags = 0;
+  }
+
+  void PrepStatX(std::string_view filepath, struct statx *stat) {
+    //AT_FDCWD is ignored when addr is an absolute path
+	  PrepFd(IORING_OP_STATX, AT_FDCWD);
```
romange, line 142: indentation

util/fibers/uring_file.h
romange: and btw, we do not have to implement all the functions possible, imho. The reason we use io_uring is for:
1. asynchronisity
2. low latency.

So if statx is being called fairly rarely - you could just use libc variant if you prefer.

util/fibers/submit_entry.h
kostasrim: Screenshot 2025-09-24 at 10 00 29 AM


It's a github bug! I explicitly set `\t` to `space` in my vim so there are `no tab characters`

PR #475, opened 2025-10-24
fix: Notify fibers on shutdown

abhijat:
A conditional wake after shutdown for epoll socket is converted to unconditional wait.

When a socket shuts down, any fibers reading from it must get a timely event to be able to react and exit their loops. If the socket is closed before this happens, the socket fd is removed from the epoll interest set, and the callback handler is reset. This can result in the reading fibers being suspended forever causing a hang.

To avoid this a wake is issued which allows such fibers to terminate their receive loop.

### DIFF
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -548,12 +548,11 @@ io::Result<size_t> EpollSocket::Recv(const io::MutableBytes& mb, int flags) {
 auto EpollSocket::Shutdown(int how) -> error_code {
   auto ec = LinuxSocketBase::Shutdown(how);
 
-#ifdef __APPLE__
-  // Since kqueue won't notify listen sockets when shutdown, explicitly wake
-  // up any read contexts. Note this will do nothing if there is no
-  // read_req_ so its safe to call multiple times.
-  Wakey(EpollProactor::EPOLL_IN, 0, nullptr);
-#endif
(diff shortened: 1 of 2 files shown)


kostasrim:
@abhijat plz sign your commits

util/fibers/epoll_socket.cc
```diff
@@ -548,12 +548,11 @@ io::Result<size_t> EpollSocket::Recv(const io::MutableBytes& mb, int flags) {
...
 
-#ifdef __APPLE__
-  // Since kqueue won't notify listen sockets when shutdown, explicitly wake
-  // up any read contexts. Note this will do nothing if there is no
-  // read_req_ so its safe to call multiple times.
+  // Wake any suspended fibers so that they can handle the shutdown event before the socket closes.
+  // In the absence of this wake if the socket closes soon after shutdown without any yields in
+  // between, suspended read fibers can hang indefinitely because socket close removes the fd from
+  // epoll interest set and unsets the callback handler.
   Wakey(EpollProactor::EPOLL_IN, 0, nullptr);
```
kostasrim: Can this happen for EPOLL_OUT as well ?

abhijat: Added `EPOLL_OUT` to the mask

kostasrim (approved):
lgtm

abhijat:
@romange @kostasrim please review, I have added the test

romange (dismissed):

romange (approved):

PR #478, opened 2025-10-30
uring: polling improvements

romange:
1. Removed epoll_array from UringProactor - now we allocate the entries on demand
2. Add multishot argument to EpollAdd that allows to use the same epoll entry for many notifications (like epoll).

### DIFF
--- a/util/fibers/uring_proactor.cc
+++ b/util/fibers/uring_proactor.cc
@@ -616,7 +616,7 @@ void UringProactor::EpollDel(EpollIndex id) {
   FiberCall fc(this);
   fc->PrepPollRemove(uid);
   IoResult res = fc.Get();
-  if (res == 0) {  // removed from iouring, othwerwise it run already and deleted itself.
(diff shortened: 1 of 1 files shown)


util/fibers/uring_proactor.cc
```diff
@@ -596,44 +595,29 @@ int UringProactor::CancelRequests(int fd, unsigned flags) {
...
+  EpollEntry* entry = reinterpret_cast<EpollEntry*>(id);
+  entry->event_mask &= ~EpollEntry::kMultishot;  // disable mask to signal no further events.
+  unsigned uid = entry->index;
 
   FiberCall fc(this);
   fc->PrepPollRemove(uid);
   IoResult res = fc.Get();
-  if (res == 0) {  // removed from iouring.
-    EpollDelInternal(id);
+  if (res == 0) {  // removed from iouring, othwerwise it run already and deleted itself.
```
kostasrim:
```suggestion
  if (res == 0) {  // removed from iouring, otherwise it run already and deleted itself.
```

PR #482, opened 2025-11-05
chore: integrate EnableRecvMultishot in uring socket with OnRecv hook

romange:
Before we had a monstrous and inefficient interface,trying to juggle recv multishot notifications together with Recv calls. This was akward, inconvenient and required multiple tracking layers.

With `OnRecv` hook, we can just pass the provided buffer to the registered callback as we receive it and immediately replenish it. Bundles processing also becomes much simpler though we still do not support them in this PR.

I have not added tests yet but checked manually via echo_server.

### DIFF
--- a/util/fiber_socket_base.h
+++ b/util/fiber_socket_base.h
@@ -129,7 +129,7 @@ class FiberSocketBase : public io::Sink,
     // For iouring/bufring notifications, it contains either the positive error code,
     // or the received data buffer. In latter case, the callback must consume the data before
     // returning.
-    std::variant<std::monostate, int, io::MutableBytes> read_result;
+    std::variant<std::monostate, std::error_code, io::MutableBytes> read_result;
   };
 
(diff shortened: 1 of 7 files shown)


kostasrim (dismissed):

util/fiber_socket_base.h
```diff
@@ -124,7 +125,11 @@ class FiberSocketBase : public io::Sink,
   virtual void CancelOnErrorCb() = 0;
 
   struct RecvNotification {
-    // TODO: move ProvidedBuffer functionality here and remove RecvProvided interface.
+    // For pure epoll notifications, this contains a monostate.
+    // For iouring/bufring notifications, it contains either the positive error code,
+    // or the received data buffer. In latter case, the callback must consume the data before
+    // returning.
+    std::variant<std::monostate, int, io::MutableBytes> read_result;
```
dranikpg: MutableBytes is already 16 bytes, so an error_code seems nicer than a tricky unix-c-stdlib negative int error code

PR #484, opened 2025-11-13
chore: Introduce TryRecv method for helio sockets

romange:
When using OnRecv interface, one may need to read data from the socket. a posix `recv` call works for regular sockets but not for tls socket that needs to delegate to SSL_read.

Also, simplify a little the async state in tls socket.

### DIFF
--- a/util/tls/tls_socket.cc
+++ b/util/tls/tls_socket.cc
@@ -492,6 +492,10 @@ io::Result<size_t> TlsSocket::TrySend(const iovec* v, uint32_t len) {
   return 0;
 }
 
+io::Result<size_t> TlsSocket::TryRecv(io::MutableBytes buf) {
+  LOG(DFATAL) << "Not implemented";
+  return 0;
+}
 
 bool TlsSocket::IsUDS() const {
   return next_sock_->IsUDS();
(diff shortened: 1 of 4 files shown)


util/tls/tls_socket.h
```diff
@@ -156,12 +156,12 @@ class TlsSocket final : public FiberSocketBase {
     enum Role : std::uint8_t { READER, WRITER };
 
     AsyncReq(TlsSocket* owner, io::AsyncProgressCb cb, const iovec* v, uint32_t len,
-             Engine::OpResult op_val, Role role)
-        : owner_(owner), caller_completion_cb_(std::move(cb)), vec_(v), len_(len), op_val_(op_val),
```
romange: removed op_val_.

util/tls/tls_socket.cc
```diff
@@ -749,27 +753,37 @@ void TlsSocket::AsyncReq::MaybeSendOutputAsync() {
...
 }
 
+/*
+   TODO: Async write path can be improved. We should separate the asynchronous flow that pulls
+   data from the engine and pushes it to the upstream socket and the flow that pushes data
+   from the user to the engine. We could call AsyncProgressCb with the result as soon as we push data to the engine, even if the engine is not flushed yet, as long as we guarantee that
+   the engine is eventually flushed. This may create cases where we "miss" socket errors,
+   as we discover them eventually. But it's fine as long as we manage this properly in tls socket
+   states. Why it is better? Because during happy path, we can push data to the engine, and then
+   flush to the socket via TrySend and all this without allocations and asynchronous state
```
kostasrim: yes!

PR #491, opened 2025-11-17
chore: do not miss notifications in Epoll::RegisterOnRecv

romange:
It could be that we get a notification and call RegisterOnRecv afterwards. Before this PR the notification would get lost.
Now it is cached so that RegisterOnRecv can call the hook.

Also, simplify error propagation in EpollSocket::Wakey.

### DIFF
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -35,6 +35,7 @@ inline EpollSocket::error_code from_errno() {
 
 #ifdef __linux__
 constexpr int kEventMask = EPOLLIN | EPOLLOUT | EPOLLET | EPOLLRDHUP;
+constexpr int kErrMask = EPOLLERR | EPOLLHUP | EPOLLRDHUP;
 
 int AcceptSock(int fd) {
   sockaddr_storage client_addr;
@@ -57,6 +58,7 @@ int CreateSockFd(int family) {
 #elif defined(__FreeBSD__) || defined(__APPLE__)
 
 constexpr int kEventMask = POLLIN | POLLOUT;
+constexpr int kErrMask = POLLERR | POLLHUP;
 
 int AcceptSock(int fd) {
   sockaddr_storage client_addr;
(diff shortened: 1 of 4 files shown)


util/fibers/epoll_socket.cc
```diff
@@ -470,66 +466,72 @@ void EpollSocket::ResetOnRecvHook() {
...
-}
 
-void EpollSocket::Wakey(uint32_t ev_mask, int error, EpollProactor* cntr) {
-  DVSOCK(2) << "Wakey " << ev_mask;
-#ifdef __linux__
-  constexpr uint32_t kErrMask = EPOLLERR | EPOLLHUP | EPOLLRDHUP;
-#else
-  constexpr uint32_t kErrMask = POLLERR | POLLHUP;
-#endif
+    bool is_send = (i == 1);
```
romange: I see diffs here are hard. I can split this PR into two commits

util/fibers/epoll_socket.cc
```diff
@@ -470,66 +466,72 @@ void EpollSocket::ResetOnRecvHook() {
...
+      };
+
+      // If we got notification, try to run the async request.
+      if (auto res = async_request->Run(native_handle(), is_send); res.first) {
+        auto cb = finalize_and_fetch_cb();
+        cb(res.second);
+      }
+    } else if (recv_hook_registered_ && !is_send) {
+      read_notified_ = 0;
+      on_recv_->cb(close_socket ? RecvNotification{make_error_code(errc::connection_aborted)}
```
kostasrim: Does that mean we missed a notification before? 

```
 else if (recv_hook_registered_ && !is_send && !ec) {
    on_recv_->cb(RecvNotification{});
```

Because `!ec` ? So we would miss and not call the on_recv_ callback?

util/fibers/epoll_socket.cc
kostasrim: It's a small change, it's ok

kostasrim (approved):

vyavdoshenko:
lgtm
I don't see anything critical, but I am not very familiar with this code.

PR #504, opened 2025-12-04
feat(tls): add async I/O support for TLS sockets 

glevkovich:
- Introduced async I/O for TlsSocket, enabling non-blocking operations on encrypted connections
- Added TrySend and TryRecv to intercept, encrypt, and decrypt data at the TLS layer
- Added RegisterOnRecv and ResetOnRecvHook for event-driven read support
- Improved error handling, state management, and code clarity (cosmetics)
- Updated test certificate path and removed obsolete code for reliability


util/tls/tls_socket.cc
```diff
@@ -196,11 +204,9 @@ io::Result<size_t> TlsSocket::RecvMsg(const msghdr& msg, int flags) {
     DCHECK(!dest.empty());
 
     Engine::OpResult op_result = engine_->Read(dest.data(), dest.size());
-
+    DVLOG(2) << "Engine::Read tried to read " << dest.size() << " bytes, got " << op_result;
```
romange: curious - why did you change this?

util/tls/tls_socket.cc
```diff
@@ -220,10 +226,16 @@ io::Result<size_t> TlsSocket::RecvMsg(const msghdr& msg, int flags) {
       continue;
     }
 
-    if (read_total > 0 && op_val == Engine::NEED_READ_AND_MAYBE_WRITE) {
-      // If we have read some data, we should not block on further reads.
-      // TODO: for async reads though we could issue a read request since we know the engine
-      // buffer is empty.
+    if ((read_total > 0) && (op_val == Engine::NEED_READ_AND_MAYBE_WRITE)) {
+      // After reading some data, the TLS engine may have generated protocol output (e.g., handshake
```
romange: a valid suggestion but why not handle it via HandleOp? instead of duplicating the code here, maybe we can call HandleOp before the condition at line 229? 

also again - I do not think it is related to TLS async and could be done in a preliminary PR with all the improvements/changes you made that are unrelated to TLS async

util/tls/tls_socket.cc
```diff
@@ -468,26 +562,161 @@ void TlsSocket::CancelOnErrorCb() {
...
 io::Result<size_t> TlsSocket::TrySend(const iovec* v, uint32_t len) {
-  LOG(DFATAL) << "Not implemented";
-  return 0;
+  size_t total_written = 0;
+  size_t iov_idx = 0;
+  size_t iov_offset = 0;
+  while (iov_idx < len) {
+    // Flush any pending output from previous iterations or previous calls.
+    while (engine_->OutputPending() > 0) {
+      auto ec = MaybeSendOutputAsync();
```
romange: based on the implementation of `MaybeSendOutputAsync` I think you have a conceptual bug here.
MaybeSendOutputAsync does not guarantee that the output buffer will be eventually flushed. This will cause short writes.

util/tls/tls_socket.cc
```diff
@@ -468,26 +562,161 @@ void TlsSocket::CancelOnErrorCb() {
...
+            return total_written;
+          }
+          // If we haven't consumed anything, return EAGAIN so the caller waits.
+          return make_unexpected(ec);
+        }
+        VLOG(1) << "MaybeSendOutputAsync failed " << ec.message();
+        return make_unexpected(ec);
+      }
+      // If MaybeSendOutputAsync succeeded, it means we wrote something.
+      // We loop to check if there is more pending output.
```
romange: yeah - this is not the way to do it.

util/tls/tls_socket.cc
```diff
@@ -468,26 +562,161 @@ void TlsSocket::CancelOnErrorCb() {
...
+        ++iov_idx;
+        iov_offset = 0;
+      }
+      // We pushed data, so we likely have output.
+      // The next iteration of the loop will handle flushing it.
+      continue;
+    }
+  }
+
+  // This loop is necessary because the TLS engine may generate additional output after each flush,
```
romange: agree with the motivation but this can not be like this. it's not the  the async behavior.

Please add a test that has socket A writing into socket B, with B not reading. Continue with writes until A.upstream will start pushing back on writes to reproduce these cases.

util/tls/tls_socket.cc
glevkovich: This is a mistake. I will revert the changes where I accidentally renamed the DVSOCK to DVLOG (probably a wrong find/replace).

util/tls/tls_socket_test.cc
```diff
@@ -217,6 +220,140 @@ TEST_P(TlsSocketTest, ShortWrite) {
...
+
+  // Register the callback to only notify a waiting fiber, but only after send_done is signaled.
+  proactor_->DispatchBrief([&] {
+    server_socket_->RegisterOnRecv([&](const util::FiberSocketBase::RecvNotification&) {
+      data_ready.Notify();
+    });
+  });
+
+  // Launch sender fiber first so send_done can be signaled before RegisterOnRecv callback waits on
+  // it.
```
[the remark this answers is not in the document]
glevkovich: changed the test, and forgot to change the comment. fixed.

util/tls/tls_socket.h
```diff
@@ -131,24 +138,36 @@ class TlsSocket final : public FiberSocketBase {
...
   error_code MaybeSendOutput();
+  error_code MaybeSendOutputAsync();
 
-  /// Read encrypted data from the network socket and feed it into the TLS engine.
-  error_code HandleUpstreamRead();
+  /// Read encrypted data from the network socket and feed it into the TLS engine's input buffer.
+  error_code ReadFromUpstreamSocket();
+  error_code ReadFromUpstreamSocketAsync();
+
+  /// Write all pending encrypted data from the TLS engine's output buffer to the network socket
```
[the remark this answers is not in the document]
glevkovich: Added a specific comment above WriteToUpstreamSocketAsync

util/tls/tls_socket.cc
```diff
@@ -71,9 +72,16 @@ auto TlsSocket::Shutdown(int how) -> error_code {
...
+  if (op_result && !state_[WRITE_IN_PROGRESS]) {
     // engine_ could send notification messages to the peer.
-    std::ignore = MaybeSendOutput();
+    // We try to send it without blocking, best effort.
+    Engine::Buffer buffer = engine_->PeekOutputBuf();
+    if (!buffer.empty()) {
+      auto res = next_sock_->TrySend(buffer);
+      if (res && (*res > 0)) {
+        engine_->ConsumeOutputBuf(*res);
+      }
```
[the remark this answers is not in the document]
glevkovich: I wanted to do a one time "best-effort" try on the shutdown, but I guess that the suggested loop is more robust and still non-blocking. Applied and fixed.

util/tls/tls_socket.cc
glevkovich: I have added the above due to breakpoint debugging, just saw that there is data  to be flushed (multiple times).

Thanks for the suggestion, I have encapsulated as suggested the logic in both HandleOp and HandleOpAsync, and I'll upload the change in the next PR, outside the TLS PR.

PR #505, opened 2025-12-05
fix: a better fix for races during the epoll cancelation flow.

romange:
As io_uring_register_sync_cancel does not guaranteee that subsequent callback won't be called, we switch to using asynchronous IORING_OP_POLL_REMOVE. We also add a leakage counter to see if our invariants hold and we indeed delete all the objects related to POLL tracking.

### DIFF
--- a/util/fibers/uring_proactor.cc
+++ b/util/fibers/uring_proactor.cc
@@ -554,7 +554,7 @@ UringProactor::EpollIndex UringProactor::EpollAdd(int fd, EpollCB cb, uint32_t e
   se.PrepPollAdd(entry->fd, event_mask);
   entry->index = se.sqe()->user_data;
   se.sqe()->len = IORING_POLL_ADD_MULTI;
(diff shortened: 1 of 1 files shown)


util/fibers/uring_proactor.cc
```diff
@@ -554,27 +565,36 @@ void UringProactor::EpollDel(EpollIndex id) {
...
+  // the call.
+
+  // When using async POLL_REMOVE we **try** to guarantee serialized order.
+  // The assumption is that callback will always run after issuing PrepPollRemove.
+  IoResult res;
+  pending_epoll_removals_++;
+
+  do {
+    FiberCall fc(this);
+    fc->PrepPollRemove(uid);
```
kostasrim: Ok after searching about this a little bit more it appears that only `io_uring_register_sync_cancel` can create orphan entries. We should not reach that state with PrepPollRemove as any other completion event related to that poll will be processed `before` the poll cancellation

PR #513, opened 2025-12-11
chore: fallocate test

kostasrim:
* add small fallocate usage test

Plz read the comments in the code (see my review)

### DIFF
--- a/util/fibers/uring_file_test.cc
+++ b/util/fibers/uring_file_test.cc
@@ -165,5 +165,46 @@ TEST_F(UringFileTest, FSync) {
   });
 }
 
+TEST_F(UringFileTest, FAllocate) {
+  string path = base::GetTestTempPath("1.log");
+  proactor_->Await([this, path] {
+    auto res = OpenLinux(path, O_RDWR | O_CREAT | O_TRUNC, 0666);
+    ASSERT_TRUE(res);
+    LinuxFile* wf = (*res).get();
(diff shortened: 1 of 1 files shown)


util/fibers/uring_file_test.cc
```diff
@@ -165,5 +165,46 @@ TEST_F(UringFileTest, FSync) {
...
+    const int mode = FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE;
+    util::fb2::FiberCall fc(proactor_.get());
+    fc->PrepFallocate(wf->fd(), mode, 0, fsize);
+    FiberCall::IoResult io_res = fc.Get();
+    ASSERT_EQ(io_res, 0);
+
+    ec = StatX(path.c_str(), &stat, wf->fd());
+    ASSERT_FALSE(ec);
+    ASSERT_EQ(stat.stx_size, 0);
+    ASSERT_EQ(stat.stx_blocks, 0);
```
kostasrim: This fires but I don't understand why 🤔 

>  const int mode = FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE;

Means that we will punch a hole on all the blocks so at least we should see a reduction yet it does not change at all 🤔 

Based on my understanding, PUCH_HOLE is not `hint` meaning that we should see a reduction in block size

PR #515, opened 2025-12-13
feat(cmake): add local tarball fallback for googletest

glevkovich:
- Download googletest release asset first, fallback to archive if needed
- Use local tarball for FetchContent to improve build reliability
- Due to recent failures downloading from archive URL, this change prioritizes the release asset URL. The archive URL is kept as a fallback mirror.

### DIFF
--- a/cmake/third_party.cmake
+++ b/cmake/third_party.cmake
@@ -202,35 +202,6 @@ FetchContent_Declare(
   URL "${GTEST_TARBALL}"
 )
 
-# # Download googletest from the release asset first, fallback to the archive mirror if the first fails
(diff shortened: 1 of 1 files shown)


cmake/third_party.cmake
```diff
@@ -165,11 +165,72 @@ function(add_third_party name)
...
+  endif()
+endif()
+
 FetchContent_Declare(
   gtest
-  URL https://github.com/google/googletest/archive/v1.15.2.tar.gz
+  URL "${GTEST_TARBALL}"
 )
 
+# # Download googletest from the release asset first, fallback to the archive mirror if the first fails
```
romange: please remove dead code

PR #520, opened 2025-12-23
fix: pass correct URI to beast::request

romange:
Current code passed scheme and the domain name (aka `https://example.com/foo?bar`) into the URI of the request, while it should have sent only the path part (aka `/foo?bar`).

Fixes most of the issue  https://github.com/dragonflydb/dragonfly/issues/6279

I found additional issues in dragonfly code when loading snapshot from path with prefix containing special characters, which will be addressed separately.

Tested manually with ovhcloud and AWS S3.

### DIFF
--- a/util/aws/http_client.cc
+++ b/util/aws/http_client.cc
@@ -76,10 +76,12 @@ std::shared_ptr<Aws::Http::HttpResponse> HttpClient::MakeRequest(
   CHECK(proactor) << "aws: http client: must run in a proactor thread";
 
   auto uri = request->GetUri();
-  std::string uri_str = uri.GetURLEncodedPath();
-  uri_str += uri.GetQueryString();  // should be prefixed with '?' and url encoded.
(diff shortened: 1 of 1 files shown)


util/aws/http_client.cc
```diff
@@ -75,12 +75,17 @@ std::shared_ptr<Aws::Http::HttpResponse> HttpClient::MakeRequest(
   ProactorBase* proactor = ProactorBase::me();
   CHECK(proactor) << "aws: http client: must run in a proactor thread";
 
+  auto uri = request->GetUri();
+  std::string uri_str = uri.GetURLEncodedPath();
+  uri_str += uri.GetQueryString();  // should be prefixed with '?' and url encoded.
```
vyavdoshenko: nit:
I would check the question mark and add it if missed.

vyavdoshenko (dismissed):

vyavdoshenko (approved):

PR #525, opened 2026-01-01
feat(fibers): Subscribe to events

dranikpg:
Extend Waiter with option to store a callable object to associavte EventCount notifications not only with fiber wakeups, but generic actions. Implement this kind of interface for BlockingCounter

### DIFF
--- a/util/fibers/detail/wait_queue.h
+++ b/util/fibers/detail/wait_queue.h
@@ -35,7 +35,7 @@ class Waiter {
     // it does not work well for slist because its reference is used by slist members
     // (probably when caching last).
     o.wait_hook.swap_nodes(wait_hook);
-    o.cntx_ = nullptr;
+    o.cntx_ = static_cast<FiberInterface*>(nullptr);
   }
 
   // safe_link is used in assertions via IsLinked() method.
@@ -72,7 +72,10 @@ class WaitQueue {
     wait_list_.erase(it);
   }
 
+  // Return true if a waiter exitsted and was notified
   bool NotifyOne(FiberInterface* active);
+
+  // Return true if any waiter was notified
   bool NotifyAll(FiberInterface* active);
 
  private:
--- a/util/fibers/synchronization.cc
+++ b/util/fibers/synchronization.cc
@@ -143,18 +143,23 @@ void EmbeddedBlockingCounter::Cancel() {
   ec_.notifyAll();
 }
 
-std::optional<EventCount::Key> EmbeddedBlockingCounter::Subscribe(detail::Waiter* w) {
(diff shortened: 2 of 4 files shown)


util/fibers/synchronization.h
```diff
@@ -4,11 +4,15 @@
...
 
+#include <util/fibers/detail/wait_queue.h>
+
 #include <atomic>
 #include <condition_variable>  // for cv_status
+#include <optional>
 
 #include "base/spinlock.h"
 #include "util/fibers/detail/fiber_interface.h"
+#include "base/logging.h"
```
romange: no need for logging.h here

dranikpg:
I don't like the interface that much but its a tradeoff between simplicity of changes/implementation and elegancy

util/fibers/detail/wait_queue.h
```diff
@@ -4,50 +4,61 @@
...
+  }
+
   // For some boost versions/distributions, the default move c'tor does not work well,
   // so we implement it explicitly.
-  Waiter(Waiter&& o) : cntx_(o.cntx_) {
+  Waiter(Waiter&& o) : cntx_{std::move(o.cntx_)} {
     // it does not work well for slist because its reference is used by slist members
     // (probably when caching last).
     o.wait_hook.swap_nodes(wait_hook);
     o.cntx_ = nullptr;
```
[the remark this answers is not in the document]
romange: it's not critical, but the line can be removed now.

util/fibers/detail/wait_queue.h
```diff
@@ -62,15 +73,13 @@ class WaitQueue {
   }
 
   bool NotifyOne(FiberInterface* active);
-  void NotifyAll(FiberInterface* active);
+  bool NotifyAll(FiberInterface* active);
```
romange: comment: what does it return?

util/fibers/synchronization.cc
```diff
@@ -143,6 +143,20 @@ void EmbeddedBlockingCounter::Cancel() {
   ec_.notifyAll();
 }
 
+std::optional<EventCount::Key> EmbeddedBlockingCounter::Subscribe(detail::Waiter* w) {
+  uint64_t v = 0;
+  while (true) {
+    auto key = ec_.prepareWait();
+    if (WaitCondition(&v)()) {
+      w->Notify(detail::FiberActive());
+      return {};
```
romange: maybe just break to avoid creating unreachable paths

romange:
@gemini-code-assist `Subscribe` name is a bit confusing as it's not clear where it refers to the `BlockingCounter` or the Waiter. Suggest other names

dranikpg:
Pushed a few fixes. It doesn't have to be perfect as we can then improve parts with more experience in using it in DF itself

PR #531, opened 2026-01-11
chore: clean-ups around iouring code

romange:
1. Fix echo_server verifiction that "echo" replies look exactly like requests. Add check-fail validation for this.
2. Clean up in iouring code, remove dead comments or code.

### DIFF
--- a/examples/echo_server.cc
+++ b/examples/echo_server.cc
@@ -12,12 +12,10 @@
 
 #include <absl/strings/escaping.h>
 #include <absl/strings/str_cat.h>
-
 #include <boost/asio/read.hpp>
 
 #include "base/histogram.h"
 #include "base/init.h"
-
 #include "util/accept_server.h"
 #include "util/asio_stream_adapter.h"
 #include "util/fibers/dns_resolve.h"
@@ -75,8 +73,7 @@ thread_local int tl_conn_id = 0;
 
 class EchoConnection : public Connection {
  public:
-  EchoConnection() {
-    id_ = tl_conn_id++;
+  EchoConnection() : id_(tl_conn_id++) {
   }
 
  private:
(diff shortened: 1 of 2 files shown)


examples/echo_server.cc
```diff
@@ -260,6 +271,32 @@ void EchoConnection::Send(bool is_raw, const vector<iovec>& vec) {
...
+          continue;
+        }
+        memcpy(seq_buf + 8 - seq_needed, v.iov_base, seq_needed);
+        v.iov_base = ((uint8_t*)v.iov_base) + seq_needed;
+        v.iov_len -= seq_needed;
+        seq_needed = 0;
+        uint64_t seq_id = absl::little_endian::Load64(seq_buf);
+        CHECK_EQ(replies_, seq_id);
+      }
+      const char* chr = (const char*)v.iov_base;
```
glevkovich: nit: use modern  c++ casting (safer):
`const char* chr = static_cast<const char*>(v.iov_base);`

examples/echo_server.cc
```diff
@@ -260,6 +271,32 @@ void EchoConnection::Send(bool is_raw, const vector<iovec>& vec) {
...
+      for (unsigned j = 0; j < v.iov_len; ++j) {
+        CHECK_EQ(chr[j], 'x');
+      }
+    }
```
glevkovich: You should probably add after the loop a critical check to ensure you actually validated the full Id in the loop, else you might accept corrupted or truncated packets:
`CHECK_EQ(seq_needed, 0) << "Response too short, did not contain full Sequence ID";`

examples/echo_server.cc
```diff
@@ -69,17 +71,20 @@ VarzCount connections("connections");
...
+    id_ = tl_conn_id++;
   }
 
  private:
   struct SendMsgState {
     bool is_raw = false;
     size_t cur_sendmsg_len = 0;
     vector<iovec> vec;
-    vector<string> kept_blobs;
+    vector<vector<char>> kept_blobs;  // must be vector<char> to be on heap.
```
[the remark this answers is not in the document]
romange: yes, to be only on heap

PR #533, opened 2026-01-13
chore: Implement support for bundles and incremental buffers.

romange:
See https://github.com/axboe/liburing/wiki/What's-new-with-io_uring-in-6.11-and-6.12 for incremental buffers explanation and https://github.com/axboe/liburing/wiki/What's-new-with-io_uring-in-6.10 for bundles.

The feature is integrated with RegisterOnRecv.

One last major thing that is left unsolved is the issue of memory management, or what happens when we are out of provided buffers.

### DIFF
--- a/util/fibers/uring_socket.cc
+++ b/util/fibers/uring_socket.cc
@@ -450,16 +450,29 @@ void UringSocket::RegisterOnRecv(OnRecvCb cb) {
       if (flags & IORING_CQE_F_BUFFER) {
         CHECK_GT(res, 0);
 
-        DCHECK(flags & IORING_CQE_F_MORE);
+        uint16_t bufid = UringProactor::BufRingIdFromFlags(flags);
+        bool incremental = (flags & IORING_CQE_F_BUF_MORE) != 0;
+        DVLOG(2) << "Received " << bufid << " " << res << " " << incremental;
 
-        // holds only for non-bundle spec.
-        DCHECK_LE(res, up->BufRingEntrySize(bufring_id));
-        uint16_t bid = flags >> IORING_CQE_BUFFER_SHIFT;
-        DVLOG(2) << "Multishot recv got buffer " << bid << " of size " << res;
+        while (res > 0) {
+          auto [buf_ptr, segment_len] = up->BufRingGetBufPtr(bufring_id, bufid, res);
+          DVLOG(2) << "Notifying " << bufid << " " << res << " " << segment_len;
+          cb(RecvNotification{io::MutableBytes(buf_ptr, segment_len)});
 
-        uint8_t* buf_ptr = up->AdvanceRecvCompletion(bufring_id, bid);
-        cb(RecvNotification{io::MutableBytes(buf_ptr, static_cast<size_t>(res))});
-        up->ReplenishBuffer(bufring_id, bid);
+          // Returns next bufid that comprises the bundle.
+          bufid = up->BufRingTrackRecvCompletion(bufring_id, segment_len, bufid, incremental);
+          res -= segment_len;
+          incremental = true;  // after the first segment in a bundle we are always incremental.
+        }
+
+        if ((flags & IORING_CQE_F_MORE) == 0) {
+          // Kernel stopped multishot recv for some reason. Most likely due to lack of buffers.
+          // We have two options here:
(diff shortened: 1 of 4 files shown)


util/fibers/uring_socket.cc
```diff
@@ -450,16 +450,29 @@ void UringSocket::RegisterOnRecv(OnRecvCb cb) {
...
+        bool incremental = (flags & IORING_CQE_F_BUF_MORE) != 0;
+        DVLOG(2) << "Received " << bufid << " " << res << " " << incremental;
+
+        while (res > 0) {
+          auto [buf_ptr, segment_len] = up->BufRingGetBufPtr(bufring_id, bufid, res);
+          DVLOG(2) << "Notifying " << bufid << " " << res << " " << segment_len;
+          cb(RecvNotification{io::MutableBytes(buf_ptr, segment_len)});
+
+          // Returns next bufid that comprises the bundle.
+          bufid = up->BufRingTrackRecvCompletion(bufring_id, segment_len, bufid, incremental);
```
kostasrim: I am a little puzzled by this and I am trying to wrap my head around it:

So, if we have an `incremental bundle` then `bool incremental = false` because `IORING_CQE_F_BUF_MORE` should not be set. Now, when we call `BufRingTrackRecvCompletion` and because `incremental = false` we will treat this bundle as non incremental. Isn't that a bug ? 

Also why do we need `incremental` at all ? We know that if a buffer group is registered as incremental then it's safe to assume it is treated as such: 

if a buffer is partially filled -> should not be advanced 
if a buffer is full -> should be advanced 

So effectively with bundles all we need to check is a) was the group registered as incremental ?(incremental_buf_ring_f_) b) if yes, advance each buffer until a partial one is found

Your comment

>  // Without bundles, it's simple - as we receive IORING_CQE_F_BUF_MORE in flags,
  // which is passed as "incremental" here. But with bundles, we do not get that flag

Should verify `incremental = false`

Am I missing something ?

romange: Lets make some order. Initially, `incremental` state was supposed to work with single buffer completions so could differentiate between incremental and non incremental so I could synchronize our internal "cached_head" and `incr_offset` and read the correct slices. 

But with bundles - it becomes more complicated, as the rule is like this: `b1, b2.. bn` all but the last one obviously filled up, so no incremental state is needed. The last one is either incremental or not - I can not know in advance so I assume incrementality. During the next notification, I check bufid and increment our state if it's not the same bufid and the kernel sent me another buffer.

So why do I still need the first part? Because bundles are sent for chunks that fill up more than one buffer or when a buffer has X capacity left, but the reply is larger than X.  It does not mean that all reads are bundles. For example a simple ping pong traffic with echo_server without pipelining will never trigger bundled CQEs. Similarly, a non-pipelined traffic of small commands won't trigger bundled traffic for Dragonfly.

Bundles are sent to save on CQEs notifications when large chunks of data are read from the socket buffer. 

You are **still correct** that in theory we could ignore `IORING_CQE_F_BUF_MORE` and `incremental` apply the same (a bit complicated) synchronization logic for single buffers and deduce "incrementality" post-factum when the next CQE comes with either the same  bufid or a different one. However it's less efficient as then we can not return the current buffer immediately to the kernel as long as we are not sure about its state, and this buffer can not be used by kernel - see `AdvanceHead()` where we not just reset the incremental state but also return the buffer back to kernel.

P.S. It could be nice that `IORING_CQE_F_BUF_MORE` would be set for the last buffer in the bundle but for some reason it's not set and I had to reverse engineer how to adjust our state machine to kernel's.  I am also not sure how much performance advantage provided buffers will give us, so far I did not see significant improvements for multi-threaded setups.

romange: >So, if we have an incremental bundle then bool incremental = false because IORING_CQE_F_BUF_MORE should not be set. Now, when we call BufRingTrackRecvCompletion and because incremental = false we will treat this bundle as non incremental. Isn't that a bug ?

no it's not because by definition bundles are sent when the provided buffer (or whatever capacity is left of it due to previous incremental state) is too small to hold the current chunk and it fills it up. Incrementality just says: you need to increase the offset of the current buffer as I am going to still use it to provide more data for the next read notification at the next offset. With bundles, all buffers but last are filled up by definition.

PR #550, opened 2026-02-24
feat(build): Use all cores for third party builds if possible

abhijat:
Allows building third parties with all cores _if explicitly requested_

* `TRDP_BUILD_USE_ALL_CORES=ON` to use all cores - opt-in
* `TRDP_BUILD_JOBS` to set specific number of jobs - defaults to 4
* by default four cores are used - same as prev. behavior

### DIFF
--- a/cmake/third_party.cmake
+++ b/cmake/third_party.cmake
@@ -60,6 +60,22 @@ endif()
 
 find_package(Threads REQUIRED)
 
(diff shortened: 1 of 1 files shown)


romange:
is it still relevant?

abhijat:
> is it still relevant?

sorry it slipped from my notice, let me change it to draft and I will review and clean up. I think its still relevant but will close if not.

PR #573, opened 2026-04-13
feat(azure): support HTTP endpoints for Azurite emulator

romange:
## Summary
- Add `IsHttps()` / `PathPrefix()` to Azure `Credentials` so the storage
  layer can connect over plain HTTP with path-style URLs (Azurite)
- `Storage::List` / `ListContainers` conditionally create SSL context
  and prepend the path prefix to all request URLs
- `OpenReadFile` / `OpenWriteFile` accept optional `ssl_cntx` and
  `path_prefix` via `ReadFileOptions`
- `gcs_demo` conditionally uses SSL based on `provider.IsHttps()`
- New Azurite-based pytest (`tests/test_azure.py`) covering list
  containers, list blobs, write, and read

## Summary by CodeRabbit

* **New Features**
  * Azure support improved: endpoint scheme (HTTP/HTTPS) is respected, SSL usage is conditional, and request path prefixes are preserved so emulator (Azurite) and real endpoints behave correctly.
  * Credential parsing exposes endpoint scheme and path-prefix for correct request construction and URL building.

* **Tests**
  * Added Azure integration tests exercising listing, write/read, prefix-aware listing, and container discovery using Azurite.

* **CI**
  * CI builds the demo binary, starts Azurite for tests, and installs Azure test dependencies.

### DIFF
--- a/util/cloud/azure/storage.h
+++ b/util/cloud/azure/storage.h
@@ -8,6 +8,7 @@
 #include <system_error>
 
 #include "io/file.h"
+#include "util/cloud/azure/creds_provider.h"
 #include "util/cloud/utils.h"
 
 typedef struct ssl_ctx_st SSL_CTX;
@@ -15,11 +16,9 @@ typedef struct ssl_ctx_st SSL_CTX;
 namespace util {
 namespace cloud::azure {
 
-class Credentials;
-
 class Storage {
  public:
-  explicit Storage(CredentialsProvider* creds) : creds_(creds) {
+  explicit Storage(Credentials* creds) : creds_(creds) {
   }
 
   using ContainerItem = std::string_view;
@@ -30,12 +29,13 @@ class Storage {
                        unsigned max_results, std::function<void(const ListItem&)> cb);
 
  private:
-  CredentialsProvider* creds_;
+  Credentials* creds_;
 };
 
 struct ReadFileOptions {
   CredentialsProvider* creds_provider = nullptr;
-  SSL_CTX* ssl_cntx;
+  SSL_CTX* ssl_cntx = nullptr;
+  std::string path_prefix;  // e.g. "/devstoreaccount1" for Azurite
 };
 
 using WriteFileOptions = ReadFileOptions;
--- a/util/cloud/azure/storage.h
+++ b/util/cloud/azure/storage.h
@@ -33,9 +33,8 @@ class Storage {
 };
 
 struct ReadFileOptions {
-  CredentialsProvider* creds_provider = nullptr;
+  Credentials* creds_provider = nullptr;
   SSL_CTX* ssl_cntx = nullptr;
-  std::string path_prefix;  // e.g. "/devstoreaccount1" for Azurite
 };
 
 using WriteFileOptions = ReadFileOptions;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
           cd build
 
           ninja -k 5 base/all io/all strings/all util/all echo_server ping_iouring_server \
-            https_client_cli s3_demo
+            https_client_cli s3_demo gcs_demo
           ls -la
 
     - name: Test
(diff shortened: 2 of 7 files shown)


romange:
Test run and pass: 
https://github.com/romange/helio/actions/runs/24343892979/job/71079481845?pr=573#step:12:70

vyavdoshenko (commented):
## Review

Overall a clean refactoring — collapsing `account_name_`, `service_endpoint_`, `is_https`, and `path_prefix` into `AccountInfo` is a good structural improvement. The Azurite CI integration test looks solid. A few observations:

### 1. Breaking change in `Storage` constructor — Dragonfly caller needs update

`Storage` now takes `Credentials*` instead of `CredentialsProvider*`. In dragonfly's `src/server/detail/snapshot_storage.cc:493` there's a call:
```cpp
cloud::azure::Storage azure(creds_provider_.get());
```
where `creds_provider_` is a `unique_ptr<CredentialsProvider>`. This will fail to compile after this change. Line 452 already had a C-style cast `(cloud::azure::Credentials*)creds_provider_.get()` to work around a similar issue — both callers need updating in the dragonfly PR.

Consider whether `Storage` truly needs `Credentials*` rather than `CredentialsProvider*`. The only reason seems to be `IsHttps()` and `PathPrefix()` — if those were exposed in the base class or via a separate config struct passed to `Storage`, we'd preserve the abstraction. But if that's intentional narrowing (Azure `Storage` only ever works with Azure `Credentials`), that's fine too — just flag the dragonfly-side breakage.

### 2. `SetCredentials` merges `SetSharedKey` + `SetSas` nicely, but `SetBearer` is left out

`SetBearer` still manually sets fields instead of delegating to `SetCredentials`. It could be:
```cpp
void Credentials::SetBearer(AccountInfo info, string token, unsigned ttl) {
  VLOG(1) << "azure: obtained access token, account=" << info.account_name
          << ", endpoint=" << info.service_endpoint << ", expires in " << ttl << " seconds";
  // Bearer doesn't use key or sas, but we could still route through SetCredentials
  // for the common fields (account_info_, source_, lock_, expire_time_ reset).
  account_info_ = std::move(info);
  ...
```
This is minor — the current code is correct, just a possible further cleanup.

### 3. `path_prefix` duplicated in `ReadFileOptions` and `Credentials`

`ReadFileOptions` has a `path_prefix` field, but `Credentials` also exposes `PathPrefix()`. This means callers must manually copy `provider.PathPrefix()` into `opts.path_prefix` (as `gcs_demo.cc` does). This is error-prone — easy to forget.

Two options:
- **Preferred**: Remove `path_prefix` from `ReadFileOptions`/`WriteFileOptions` and have `OpenReadFile`/`OpenWriteFile`/`WriteFile` read it from the `CredentialsProvider` (already available via `opts.creds_provider`). This requires a downcast or adding `PathPrefix()` to the base class.
- **Alternative**: Keep as-is but document that callers must set it.

### 4. `WriteFile` constructor takes `opts` by value — good, but `ReadFile` gets a copy too

`ReadFile` constructor now moves `opts`:
```cpp
ReadFile(string read_obj_url, ReadFileOptions opts)
    : read_obj_url_(std::move(read_obj_url)), opts_(std::move(opts)) {
```
Good. But `OpenReadFile` receives `const ReadFileOptions&`, so there'll be a copy into the `ReadFile` constructor anyway. Not a big deal for this struct, just noting the inconsistency.

### 5. CI: Node.js version pinned in Azurite step

```yaml
curl -fsSL https://nodejs.org/dist/v22.16.0/node-v22.16.0-linux-x64.tar.xz | tar -xJ -C /tmp
```
This will break when the Node.js version is removed from their CDN. Consider using `actions/setup-node` or a version range instead. Minor — CI-only.

### 6. Missing newline at EOF fixed — nice

The missing newline at the end of several files was fixed as part of this PR. 👍

### Summary
The core changes (conditional SSL, path prefix propagation) are correct and well-tested. The main action items are:
1. Coordinate the dragonfly-side `Storage` constructor update
2. Consider collapsing `path_prefix` to avoid duplication between opts and credentials

util/cloud/azure/storage.h
```diff
@@ -30,12 +29,13 @@ class Storage {
...
  private:
-  CredentialsProvider* creds_;
+  Credentials* creds_;
 };
 
 struct ReadFileOptions {
   CredentialsProvider* creds_provider = nullptr;
-  SSL_CTX* ssl_cntx;
+  SSL_CTX* ssl_cntx = nullptr;
+  std::string path_prefix;  // e.g. "/devstoreaccount1" for Azurite
```
vyavdoshenko: nit
Is this the same path_prefix as in AccountInfo? Can it be reused?

vyavdoshenko (dismissed):

util/cloud/azure/storage.h
romange: good catch. removed.

vyavdoshenko (approved):

PR #576, opened 2026-04-16
feat: stack overflow safeguard for dispatch fiber

kostasrim:
This PR adds two safeguards:

- sigaltstack + SIGSEGV handler (epoll_proactor.cc): each proactor thread registers a 64KB alternate signal stack and a SA_ONSTACK | SA_SIGINFO handler. When a stack overflow triggers a fault, the handler reads the interrupted SP from ucontext_t, compares it against the active fiber's stack_bottom_, and if it has overflowed writes a clear message to stderr and aborts _ before the corrupted heap produces a misleading crash.

- InitStackBottom for the dispatcher fiber (scheduler.cc): the dispatcher was the only fiber that never called InitStackBottom, leaving stack_bottom_ null and making overflow detection impossible for it. Now it paints canaries and records its stack bounds the same way worker fibers do.

This won't catch every case _ by the time SIGSEGV fires the heap may already be partially corrupted _ but it reliably produces an actionable message instead of a 285-frame backtrace ending in for example `std::string::_M_dispose` like seen in: https://github.com/dragonflydb/dragonfly/issues/3464

## Summary by CodeRabbit

* **Bug Fixes**
  * Dispatcher initialization now sets stack canaries for its preallocated stack so runtime stack-safety checks behave consistently with worker fibers.
  * Added platform-specific signal diagnostics (Linux/FreeBSD only; disabled on macOS and when sanitizers are active) to detect and report probable fiber stack overflows.
  * Establishes a per-thread alternate signal stack and enables the overflow handler only on the primary proactor thread for reliable diagnostics.

### DIFF
--- a/util/fibers/detail/scheduler.cc
+++ b/util/fibers/detail/scheduler.cc
@@ -74,6 +74,9 @@ DispatcherImpl::DispatcherImpl(const ctx::preallocated& palloc, ctx::fixedsize_s
                                detail::Scheduler* sched) noexcept
     : FiberInterface{DISPATCH, FiberPriority::NORMAL, 0, "_dispatch"} {
   stack_size_ = palloc.sctx.size;
+  // Paint canaries before constructing the fiber context so CheckStackMargin works
+  // for the dispatcher the same way it does for worker fibers.
+  InitStackBottom(reinterpret_cast<uint8_t*>(palloc.sp) - palloc.size, palloc.size);
   entry_ = ctx::fiber(std::allocator_arg, palloc, std::move(salloc),
                       [this](ctx::fiber&& caller) { return Run(std::move(caller)); });
   scheduler_ = sched;
--- a/util/fibers/epoll_proactor.cc
+++ b/util/fibers/epoll_proactor.cc
@@ -96,7 +96,7 @@ void OnSigSegv(int sig, siginfo_t* info, void* ucontext_arg) {
   // skip the check rather than producing a false-positive fatal error.
   if (sp && stack_bottom && sp < stack_bottom) {
     const char msg[] = "POSSIBLE FIBER STACK OVERFLOW DETECTED\n";
-    write(STDERR_FILENO, msg, sizeof(msg) - 1);
+    std::ignore = write(STDERR_FILENO, msg, sizeof(msg) - 1);
   }
 
   // Restore the previous handler (e.g. Abseil's) and return. The kernel re-executes the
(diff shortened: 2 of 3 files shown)


kostasrim:
I will send a fix to the epoll socket soon. This PR will help us identify these kind of long recursions

util/fibers/detail/scheduler.cc
```diff
@@ -74,6 +74,9 @@ DispatcherImpl::DispatcherImpl(const ctx::preallocated& palloc, ctx::fixedsize_s
                                detail::Scheduler* sched) noexcept
     : FiberInterface{DISPATCH, FiberPriority::NORMAL, 0, "_dispatch"} {
   stack_size_ = palloc.sctx.size;
+  // Paint canaries before constructing the fiber context so CheckStackMargin works
```
romange, line 77: what's this "Paint canaries" ?

util/fibers/epoll_proactor.cc
```diff
@@ -46,6 +48,48 @@ namespace {
...
+    write(STDERR_FILENO, suffix, sizeof(suffix) - 1);
+    abort();
+  }
+
+  // Not a stack overflow — restore default handler and re-raise to produce a core dump.
+  signal(SIGSEGV, SIG_DFL);
+  raise(SIGSEGV);
+}
+
+void InstallSigSegvHandler() {
```
romange, line 84: frankly, I do not understand this well enough and I am hesitant to accept it.

1. it fails unit tests
2. what happens when a regular segfault happens? We are loosing our stacktrace context?

kostasrim: I am aware for (1). Will explain for (2)

kostasrim: > what happens when a regular segfault happens? We are loosing our stacktrace context?

Nope, abseil handlers should kick in which in turn should print the stack trace of the fiber that segfaulted. Based on my understanding, signals work in chains, we add some wrapper, do some work and then pass control (or exit) to the previous handler in the chain.

Writing this made me realize we do not need to abort() or fail in case we detect that the fiber stack overflowed. We can just add a WRITE() that just advertises the possibility of overflow and then crash as we normally would with the "missleading" stack trace. That way we don't break (or worry) any of the existing functionality.

Note, that even if we aborted() in the new handler we could use (or write) any tooling that can traverse the ucontext directly (since we have access to it) which might have its use cases (similar to what abseil does when we segfault and we get the stack trace)

Plz hold off the review. I will re-request when ready.

romange: please manually check these cases: check-fail, segfault, and stack-overflow - before and after the change and make sure stack-traces do not regress for  any of these cases.

kostasrim: +1

PR #577, opened 2026-04-16
fix: epoll async read/write can stack overflow the dispatch fiber

kostasrim:
Epoll-backed AsyncWriteSome and AsyncReadSome fire their completion callbacks inline when the operation succeeds immediately _ unlike io_uring which always defers through the submission ring. This is an intentional optimization: avoiding the round-trip through the ring reduces latency for sockets that are immediately readable/writable.

However, if a completion callback re-issues an async operation that also completes immediately (e.g. JournalStreamer::OnCompletion _ AsyncWrite _ AsyncWriteSome _ TrySend succeeds), the chain recurses entirely on the dispatcher fiber's 32KB stack. With enough pending buffers this exhausts the stack, silently corrupts adjacent heap memory, and produces a crash with no actionable diagnostic.

This PR fixes this case by applying the cb via DispatchBrief which breaks the cycle/recursion. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed callback timing for asynchronous writes: completion callbacks are now always dispatched through the async scheduler instead of sometimes running inline, improving stability under high concurrency, reducing risk of stack overflows, and ensuring more consistent callback ordering.

### DIFF
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -365,7 +365,20 @@ void EpollSocket::AsyncWriteSome(const iovec* v, uint32_t len, io::AsyncProgress
   Result<size_t> res = TrySend(v, len);
 
   if (res || res.error().value() != EAGAIN) {
-    cb(res);
+    // tl_async_cb_depth guards against unbounded synchronous recursion. When TrySend succeeds
+    // without EAGAIN the callback fires inline — if that callback issues another AsyncWrite
+    // which also succeeds inline, the chain recurses entirely on the dispatcher fiber's stack
+    // (32KB) until it overflows. Deferring via DispatchBrief when already inside a completion
+    // callback breaks the chain: the next iteration of the proactor event loop picks it up.
+    thread_local unsigned tl_async_cb_depth = 0;
+    if (tl_async_cb_depth < 5) {
+      ++tl_async_cb_depth;
+      cb(res);
+      --tl_async_cb_depth;
+    } else {
+      auto* proactor = GetProactor();
+      proactor->DispatchBrief([cb = std::move(cb), res]() mutable { cb(res); });
+    }
     return;
   }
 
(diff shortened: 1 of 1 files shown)


util/fibers/epoll_socket.cc
```diff
@@ -365,7 +365,20 @@ void EpollSocket::AsyncWriteSome(const iovec* v, uint32_t len, io::AsyncProgress
...
 
   if (res || res.error().value() != EAGAIN) {
-    cb(res);
+    // tl_async_cb_depth guards against unbounded synchronous recursion. When TrySend succeeds
+    // without EAGAIN the callback fires inline — if that callback issues another AsyncWrite
+    // which also succeeds inline, the chain recurses entirely on the dispatcher fiber's stack
+    // (32KB) until it overflows. Deferring via DispatchBrief when already inside a completion
+    // callback breaks the chain: the next iteration of the proactor event loop picks it up.
+    thread_local unsigned tl_async_cb_depth = 0;
+    if (tl_async_cb_depth < 5) {
```
kostasrim, line 374: 5 is a heuristic not a panacea. I do not want to penaltize this optmization but the true value here depends on the stack size and its contents 🤷

util/fibers/epoll_socket.cc
```diff
@@ -365,7 +365,20 @@ void EpollSocket::AsyncWriteSome(const iovec* v, uint32_t len, io::AsyncProgress
...
+      ++tl_async_cb_depth;
+      cb(res);
+      --tl_async_cb_depth;
+    } else {
+      auto* proactor = GetProactor();
```
kostasrim, line 379: Wtf did the AI do 🤣 Let me fix that quickly

kostasrim: The problem here is that if it falls through it won't work as the notifications here are edge triggered :/ 

Let me think 🤔

util/fibers/epoll_socket.cc
```diff
@@ -365,7 +365,20 @@ void EpollSocket::AsyncWriteSome(const iovec* v, uint32_t len, io::AsyncProgress
   Result<size_t> res = TrySend(v, len);
 
   if (res || res.error().value() != EAGAIN) {
-    cb(res);
+    // tl_async_cb_depth guards against unbounded synchronous recursion. When TrySend succeeds
```
romange: I would like to simplify this. We run io_uring and are quite happy with its performance. 
Therefore I think we can afford calling DispatchBrief unconditionally. 

There is also the issue of DispatchBrief possibly blocking if the dispatch queue is full - a very rate condition but it will cause a deadlock.

romange: ah, sorry we are not running from the callback. we can call DispatchBrief and block.
so lets call `DispatchBrief` unconditionally here.

romange: No, actually we might have a problem here with DispatchBrief as `AsyncWriteSome` may be called from a callback being run in the dispatch fiber.

So `AsyncWriteSome` is called, its completion callback is called from the dispatch fiber, it calls `AsyncWriteSome` again, which calls DispatchBrief that may block and cause deadlocks.

it's a design problem that we can not fix now. Please add a TODO here, flagging this scenario.

kostasrim: yes, any async cb should not preempt as it does not run in fiber context. I wrote a comment, will push in a few mins

kostasrim: done

kostasrim: also async read is fiber blocking -- will revert the change.

romange (dismissed):

romange (approved):

PR #578, opened 2026-04-19
feat(cloud): add cursor pagination to S3/GCS/Azure List

romange:
## Summary
- Extend `S3Storage::List`, `GCS::List`, and `azure::Storage::List` with a required `continuation_token` in/out pointer. Each call performs a single request: the token is read on input (empty = start from the beginning) and overwritten with the next page's token on output, or cleared when the last page is reached. Callers drive pagination by looping until the token comes back empty.
- `GCS::List` gains a `max_results` parameter (was hardcoded to 200), matching S3 and Azure.
- Azure's `List` now parses `<NextMarker>` — previously it only returned the first page.
- `examples/s3_demo.cc` drives `list-objects` via the cursor loop and prints `-- page N end --` markers so pagination can be exercised from tests. `examples/gcs_demo.cc` loops externally for the GCS and Azure listings.
- New `test_list_objects_cursor_pagination` in `tests/test_s3.py` uploads 5 objects, lists with page size 2 against MinIO, and asserts every key plus 3 page markers appear in the output.

## Test plan
- [x] `ninja s3_demo gcs_demo` (build-dbg)
- [x] `pytest tests/test_s3.py -v` against local MinIO — `test_put_get_object_minio` and `test_list_objects_cursor_pagination` pass
- [x] Manual end-to-end: `s3_demo --cmd=list-objects --list_max_results=2` on 5 objects produces 3 pages (2/2/1)
- [ ] CI green against real S3/MinIO

## Summary by CodeRabbit

* **New Features**
  * Listings across cloud providers now use caller-driven, single-page pagination with configurable page sizes, explicit per-page end markers during multi-page listings, and consistent resume-token handling.

* **Bug Fixes**
  * Improved pagination reliability and immediate error reporting when page requests fail.

* **Tests**
  * Added an integration test that verifies cursor-based pagination, page-size behavior, and per-page end markers.

### DIFF
--- a/examples/gcs_demo.cc
+++ b/examples/gcs_demo.cc
@@ -105,11 +105,11 @@ void Run(SSL_CTX* ctx) {
         cout << "Object: " << item.key << ", size: " << item.size
              << " mtime: " << absl::FromUnixNanos(item.mtime_ns) << endl;
       };
-      string token;
+      string cursor;
       do {
-        ec = gcs.List(bucket, prefix, false, 200, cb, &token);
+        ec = gcs.List(bucket, prefix, false, 200, cb, &cursor);
         if (ec) break;
-      } while (!token.empty());
+      } while (!cursor.empty());
     }
   } else {
     auto cb = [](std::string_view bname) { CONSOLE_INFO << bname; };
@@ -178,11 +178,11 @@ void RunAzure(SSL_CTX* ctx) {
     auto cb = [](const util::cloud::azure::Storage::ListItem& item) {
       CONSOLE_INFO << "Object: " << item.key << " " << item.size << " " << item.mtime_ns << endl;
     };
-    string token;
+    string cursor;
     do {
-      ec = storage.List(bucket, prefix, false, 100, cb, &token);
+      ec = storage.List(bucket, prefix, false, 100, cb, &cursor);
       if (ec) break;
-    } while (!token.empty());
+    } while (!cursor.empty());
     LOG_IF(ERROR, ec) << "Error listing " << bucket << " " << ec.message();
   }
 }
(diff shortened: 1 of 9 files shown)


examples/gcs_demo.cc
```diff
@@ -105,7 +105,11 @@ void Run(SSL_CTX* ctx) {
         cout << "Object: " << item.key << ", size: " << item.size
              << " mtime: " << absl::FromUnixNanos(item.mtime_ns) << endl;
       };
-      ec = gcs.List(bucket, prefix, false, cb);
+      string token;
+      do {
+        ec = gcs.List(bucket, prefix, false, 200, cb, &token);
+        if (ec) break;
+      } while (!token.empty());
```
BorysTheDev: while (!token.empty()); looks strange for request, could you add some comments

examples/gcs_demo.cc
```diff
@@ -171,11 +175,14 @@ void RunAzure(SSL_CTX* ctx) {
...
-                                     << item.mtime_ns << endl;
-                      });
+    auto cb = [](const util::cloud::azure::Storage::ListItem& item) {
+      CONSOLE_INFO << "Object: " << item.key << " " << item.size << " " << item.mtime_ns << endl;
+    };
+    string token;
+    do {
+      ec = storage.List(bucket, prefix, false, 100, cb, &token);
+      if (ec) break;
+    } while (!token.empty());
```
BorysTheDev: the same

examples/gcs_demo.cc
BorysTheDev: or rename to cursor

BorysTheDev (approved):

examples/gcs_demo.cc
romange: @gemini-code-assist rename token to "cursor"

BorysTheDev (approved):

PR #579, opened 2026-04-23
test(fibers): add DispatchLocalBrief test

romange:
## Summary

- Adds `ProactorBase::DispatchLocalBrief`, a fire-and-forget dispatch that must run on the proactor's own thread and never blocks: when the task queue is full, it drains queued tasklets until there is room to enqueue the new one.
- Adds `ProactorTest.DispatchLocalBrief` covering all four proactor/IP variants (epoll/uring × IPv4/IPv6), submitting `2 × kTaskQueueLen` tasks to exercise the self-drain path with an `EventCount` sentinel for completion.
- Tidies `DispatchBrief`'s doc comment (`//!` → `//`) and documents its return value semantics.

## Changes

- `util/fibers/proactor_base.h`: adds `DispatchLocalBrief` declaration and template implementation; updates `DispatchBrief` doc comment.
- `util/fibers/fibers_test.cc`: adds `ProactorTest.DispatchLocalBrief`.

## Test plan

- [x] `build-dbg/fibers_test --gtest_filter="*DispatchLocalBrief*"` — all 4 variants pass

## Summary by CodeRabbit

* **New Features**
  * Added a new task dispatch method with integrated queue management and automatic overflow handling for scheduling work on the proactor thread.

* **Tests**
  * Added test coverage for the new dispatch method, validating concurrent task execution and synchronization behavior.

* **Documentation**
  * Updated dispatch method documentation to clarify return semantics and behavior expectations.

### DIFF
--- a/util/fibers/proactor_base.h
+++ b/util/fibers/proactor_base.h
@@ -116,10 +116,17 @@ class ProactorBase {
     return task_queue_.is_full();
   }
 
-  //! Fire and forget - does not wait for the function to run called.
-  //! `f` should not block, lock on mutexes or Await.
-  //! Might block the calling fiber if the queue is full.
+  // Fire and forget - does not wait for the function to run called.
+  // `f` should not block, lock on mutexes or Await.
+  // Might block the calling fiber if the queue is full.
+  // Returns false if the task was enqueued without waiting, true otherwise.
   template <typename Func> bool DispatchBrief(Func&& brief);
+
+  // DispatchLocalBrief is similar to DispatchBrief but is used to dispatch tasks asynchronously
+  // to the same proactor thread.
+  // Unlike DispatchBrief, it never blocks, and if the task queue is full,
+  // it will unload the queue by fetching and running tasks until there is space in the queue.
+  template <typename Func> void DispatchLocalBrief(Func&& brief);
 
   //! Similarly to DispatchBrief but 'f' is wrapped in fiber.
   //! f is allowed to fiber-block or await.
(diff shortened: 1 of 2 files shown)


util/fibers/proactor_base.h
```diff
@@ -116,11 +116,18 @@ class ProactorBase {
...
+  // `f` should not block, lock on mutexes or Await.
+  // Might block the calling fiber if the queue is full.
+  // Returns false if the task was enqueued without waiting, true otherwise.
   template <typename Func> bool DispatchBrief(Func&& brief);
 
+  // DispatchLocalBrief is similar to DispatchBrief but is used to dispatch tasks asynchronously
+  // to the same proactor thread.
+  // Unlike DispatchBrief, it never blocks, and if the task queue is full,
+  // it will unload the queue by fetching and running tasks until there is space in the queue.
+  template <typename Func> void DispatchLocalBrief(Func&& brief);
```
kostasrim, line 129: Solves https://github.com/romange/helio/pull/577/changes#diff-7f07ac3b33b79ef9ca64edf1ca113fe296ac5438e7d1e7f0856be35a6f005080R370

kostasrim (approved):
Will work for the epoll changes 🚢 🇮🇹

util/fibers/proactor_base.h
```diff
@@ -396,6 +403,19 @@ template <typename Func> bool ProactorBase::DispatchBrief(Func&& f) {
...
 }
 
+template <typename Func> void ProactorBase::DispatchLocalBrief(Func&& brief) {
+  assert(InMyThread());
+  while (true) {
+    if (EmplaceTaskQueue(std::forward<Func>(brief)))
+      return;
+
+    Tasklet task;
+    if (task_queue_.try_dequeue(task)) {
```
kostasrim, line 413: Something that popped to my mind, if for some reason the `task` enqueues this can spin for a lot of time or even create an endless loop. Maybe worth a tsc clock timer here to warn against these paths ? Should also be able to reproduce with your test, if the tasks use DispatchLocalBrief

kostasrim: The livelock is actually easy to break:

1. Fetch X elements in a buffer and out of the queue
2. Enque the new task
3. Now drain the buffer (for task in buffer run task() )

This should break the cycle but I didn't ask claude to verify.

romange: `task` can not enqueue as `task` should not preempt and this is a blocking queue. Blocking on it creates a deadlock, which is why we are introducing this function in the first place.

PR #580, opened 2026-04-26
fix(fibers): break-cb leaks under sustained accept+close churn

romange:
## Summary

Fix two single-shot error-callback bugs and add focused regressions for them.

### 1. `UringSocket::CancelOnErrorCb()` could cancel the wrong poll

`RegisterOnErrorCb()` stores the poll-add SQE `user_data` in `error_cb_id`. Once the CQE fires, that completion-entry slot can be immediately reused by another submission on the same proactor. If `CancelOnErrorCb()` later issued `PrepPollRemove(error_cb_id)` using the stale id, it could remove a different socket's newly armed poll request.

**Fix:**
- mark the wrapper as no longer armed by invalidating `error_cb_id` in the completion lambda
- skip `PrepPollRemove` in `CancelOnErrorCb()` once the wrapper is no longer armed
- require a non-empty error callback and expose an explicit armed-state helper on `ErrorCbRefWrapper`

### 2. `EpollSocket::Wakey()` could invoke the same error callback twice

A follow-up edge such as `POLLHUP` after an initial `POLLRDHUP` could invoke the same registered callback again on the same fd.

**Fix:**
- make `Wakey()` consume `error_cb_` before invoking it, making the registration effectively single-shot

## Tests

Added focused regressions in `util/fibers/fiber_socket_test.cc`:

- `PollSingleShot`
  - reproduces the epoll duplicate-callback path with a staged peer shutdown / close sequence
  - passes on the fixed code and is skipped for IPv6

- `PollCancelSlotReuse`
  - reproduces the uring stale-slot removal bug with a deterministic two-connection sequence:
    1. arm A
    2. fire A
    3. arm B so it reuses A's freed slot
    4. cancel A
  - verifies B still receives its callback

The socket test harness was also tightened to support these targeted regressions via `LaunchAcceptFiber()` and `ConnectClient()` helpers.

## Validation

Ran:

- `./fiber_socket_test --gtest_filter='*PollSingleShot*:*PollCancelSlotReuse*'`

Result:
- passing variants: `epoll_IPv4`, `uring_IPv4`, and all `PollCancelSlotReuse` parameterizations
- expected skips: IPv6 variants for `PollSingleShot`

## Files

- `util/fibers/epoll_socket.cc`
- `util/fibers/uring_socket.{h,cc}`
- `util/fibers/fiber_socket_test.cc`

## Summary by CodeRabbit

* **New Features**
  * Added simplified API overloads for cloud storage listing (AWS S3, Azure, GCP) that handle pagination internally without requiring explicit continuation/page tokens.

* **Bug Fixes**
  * Fixed error callbacks in socket handling to fire only once, preventing duplicate error notifications from the same socket error.
  * Corrected stale slot reuse issue in uring socket error callback cancellation logic.

* **Tests**
  * Added regression tests for socket error handling edge cases and callback lifecycle validation.

### DIFF
--- a/util/fibers/uring_socket.cc
+++ b/util/fibers/uring_socket.cc
@@ -368,7 +368,8 @@ io::Result<size_t> UringSocket::Recv(const io::MutableBytes& mb, int flags) {
 
     if (res >= 0) {
       has_recv_data_ = (fc.flags() & IORING_CQE_F_SOCK_NONEMPTY) ? 1 : 0;
-      DVSOCK(2) << "Received " << res << " bytes " << ", has_more " << has_recv_data_;
+      DVSOCK(2) << "Received " << res << " bytes "
+                << ", has_more " << has_recv_data_;
       return res;
     }
     DVSOCK(2) << "Got " << res;
@@ -391,6 +392,7 @@ io::Result<size_t> UringSocket::Recv(const io::MutableBytes& mb, int flags) {
 void UringSocket::RegisterOnErrorCb(std::function<void(uint32_t)> cb) {
   CHECK(!error_cb_wrapper_);
   DCHECK(IsOpen());
+  DCHECK(cb);
 
   uint32_t event_mask = POLLERR | POLLHUP;
 
@@ -398,6 +400,10 @@ void UringSocket::RegisterOnErrorCb(std::function<void(uint32_t)> cb) {
   error_cb_wrapper_ = ErrorCbRefWrapper::New(std::move(cb));
   auto se_cb = [data = error_cb_wrapper_](detail::FiberInterface*, Proactor::IoResult res,
                                           uint32_t flags, uint32_t) {
+    // Mark the poll id invalid before Destroy so a subsequent CancelOnErrorCb
+    // on the socket skips PrepPollRemove on what may now be a reused
+    // centries_ slot.
+    data->error_cb_id = ErrorCbRefWrapper::kInvalidErrorCbId;
     auto cb = std::move(data->cb);
     ErrorCbRefWrapper::Destroy(data);
     if (res < 0) {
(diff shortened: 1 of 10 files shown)


romange:
@coderabbitai review

util/fibers/uring_socket.cc
```diff
@@ -411,13 +417,25 @@ void UringSocket::RegisterOnErrorCb(std::function<void(uint32_t)> cb) {
...
 
 void UringSocket::CancelOnErrorCb() {
   if (!error_cb_wrapper_)
     return;
 
+  // If the error-poll completion already fired, the centries_ slot identified
+  // by error_cb_id has been freed and may have been reused by a subsequent
+  // submission on this proactor. Issuing PrepPollRemove on that user_data now
+  // would cancel an unrelated request. Just release our reference.
+  if (!error_cb_wrapper_->IsArmed()) {
```
kostasrim, line 433: I think I understand but plz confirm:

1. RegisterOnErrorCb. -> Submits a POLL
2. socket unwrittable (peer disconnects) -> POLLHUP
3. Now we process POLLHUP which resolves (1), the poll submitted. That frees its centry (let's call it X)
4. CancelOnErrorCb on teardown now references X which might have been reused

kostasrim: so the gist of it is by the time CancelOnErrorCb fires, X is stale

kostasrim (approved):

util/fibers/uring_socket.cc
romange: yes, CQE slots are being reused.

kostasrim: uh so it required load to manifest -> exchaust the slots so we end up reusing them by the time we Cancel. This BROKE the contract/invariant

PR #590, opened 2026-05-26
fix(socket): graceful Shutdown on invalid fd

### LINKED ISSUE #7410: SIGABRT crash in LinuxSocketBase::Shutdown during failed S3 snapshot upload
## Summary

If an S3-compatible backend (like MinIO) rejects a snapshot upload (e.g., due to being out of disk space) and returns a non-2xx XML error without an `ETag` header, the Dragonfly server crashes with a `SIGABRT`.

A failed snapshot upload should gracefully abort the `SAVE` operation and return an `-ERR` to the client, rather than crashing the entire database process.

## Reproduction / Trigger

This was caught in CI when the local MinIO container ran out of disk space. MinIO rejected the upload with a `507 Insufficient Storage` (`XMinioStorageFull`) error.

Because the response was an XML error rather than a successful upload confirmation, it lacked an `ETag`. The AWS client logs this missing ETag and attempts to clean up. Furthermore, MinIO includes `Connection: close` in the error response, which forces the client to shut down the socket.

**Dragonfly Logs:**

```text
E0525 16:27:47.279848   16673 s3_storage.cc:344] aws: UploadPart 1: missing ETag in response

```

## Root Cause & Stack Trace

When the upload fails and the server sends `Connection: close`, the `SenderResult` goes out of scope during the `Upload()` error handling. This triggers the destructor for the HTTP client (`~SenderResult()` -> `Client::Shutdown()`).

`Client::Shutdown()` calls down into the `helio` network library's `LinuxSocketBase::Shutdown(int)` at `fiber_socket_base.cc:184`. Because the socket was already invalidated or never fully established for the payload, the file descriptor (`fd_`) is `-1`.

The `helio` library has a hardcoded assertion checking `fd_ >= 0` (`Check failed: fd_ >= 0 (-1 vs. 0)`). When it sees `-1`, it raises Signal 6 (SIGABRT).

**Stack Trace (from CI):**

```text
➜    @            0xef7dec  util::LinuxSocketBase::Shutdown()
➜    @            0xed9d10  util::http::Client::Shutdown()
➜    @            0xe0d1a8  util::cloud::RobustSender::SenderResult::~SenderResult()
➜    @            0xdf4718  util::cloud::aws::(anonymous namespace)::WriteFile::Upload()
➜    @            0xdf4f74  util::cloud::aws::(anonymous namespace)::WriteFile::Close()

```

## Expected Behavior / Proposed Fix

We need to gracefully handle invalid sockets during shutdown. Possible fixes:

1. Guard `Client::Shutdown()` in `http_client.cc` to check `socket_->IsOpen()` before attempting to call `Shutdown()`.
2. Modify `LinuxSocketBase::Shutdown()` to return an error `std::error_code` instead of `CHECK`-failing when `fd_ < 0`.

*(Note: There is also an inverted `LOG_IF(WARNING, !ec)` condition in `Client::Shutdown` that logs on success rather than failure, which should likely be fixed alongside this).*

CI run: 
https://github.com/dragonflydb/dragonfly/actions/runs/26409102411/job/77739332775

Logs:

[logs (1).zip](https://github.com/user-attachments/files/28232793/logs.1.zip)

romange:
## Summary
- `LinuxSocketBase::Shutdown` no longer SIGABRTs when `fd_ < 0`; it returns `bad_file_descriptor` (DCHECK retained for debug builds).
- `Client::Shutdown` guards with `socket_->IsOpen()` so a shutdown on an already-invalidated HTTP client socket is a no-op.
- Fixes two inverted `LOG_IF(WARNING, !ec)` conditions in `http_client.cc` that logged on success rather than on error.

Fixes dragonflydb/dragonfly#7410 — the S3 upload error path destroys a `SenderResult` whose embedded HTTP client calls `Shutdown` on a socket with `fd_ == -1` (e.g. after MinIO returns `507` with `Connection: close`), previously crashing the server.

## Test plan
- [x] Existing socket / http client tests pass
- [x] Manually reproduce the S3 error path against MinIO and confirm SAVE returns `-ERR` instead of crashing

## Summary by CodeRabbit

* **Bug Fixes**
  * Treat invalid/closed sockets as a normal runtime error instead of relying only on debug assertions.
  * Only attempt socket shutdown when a socket exists and is open; improved warning logging when shutdown/close fail.
  * Preserve and return the most relevant retry-related errors when send attempts encounter connection closure, server pushback, or auth refresh.

* **Tests**
  * Added regression tests that simulate a failing storage endpoint to ensure uploads fail gracefully without crashing.

* **Documentation**
  * Clarified behavior of the robust sender in its API comments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/romange/helio/pull/590?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -172,90 +172,120 @@ def test_endpoint_flag_override(minio):
     )
 
 
-class _FailingUploadHandler(http.server.BaseHTTPRequestHandler):
-    """Fake S3 endpoint: accepts CreateMultipartUpload, fails every UploadPart
-    with 507 + `Connection: close`, mirroring MinIO's behavior when out of disk.
-    Reproduces dragonflydb/dragonfly#7410."""
+def _make_failing_upload_handler(*, connection_close: bool):
+    """Build an http handler that accepts CreateMultipartUpload but answers
+    every UploadPart with 507. `connection_close=True` mirrors MinIO's
+    XMinioStorageFull response (drives the Shutdown crash path).
+    `connection_close=False` keeps the socket alive so all retries hit
+    DoesServerPushback (drives the RobustSender retry-exhaustion path)."""
 
-    def log_message(self, *args, **kwargs):  # silence per-request stderr noise
-        pass
+    class Handler(http.server.BaseHTTPRequestHandler):
+        protocol_version = "HTTP/1.1"
 
-    def _send(self, status: int, body: bytes, *, close: bool = False):
-        self.send_response(status)
-        self.send_header("Content-Type", "application/xml")
-        self.send_header("Content-Length", str(len(body)))
-        if close:
-            self.send_header("Connection", "close")
-            self.close_connection = True
-        self.end_headers()
-        self.wfile.write(body)
+        def log_message(self, *args, **kwargs):
+            pass
 
-    def do_POST(self):
-        query = parse_qs(urlparse(self.path).query, keep_blank_values=True)
-        if "uploads" in query:
+        def _send(self, status: int, body: bytes, *, close: bool):
+            self.send_response(status)
+            self.send_header("Content-Type", "application/xml")
+            self.send_header("Content-Length", str(len(body)))
+            if close:
+                self.send_header("Connection", "close")
+                self.close_connection = True
+            else:
+                self.send_header("Connection", "keep-alive")
+            self.end_headers()
+            self.wfile.write(body)
+
+        def do_POST(self):
+            query = parse_qs(urlparse(self.path).query, keep_blank_values=True)
+            if "uploads" in query:
+                body = (
+                    b'<?xml version="1.0" encoding="UTF-8"?>'
+                    b"<InitiateMultipartUploadResult>"
+                    b"<Bucket>fake</Bucket><Key>fake</Key>"
+                    b"<UploadId>fake-upload-id</UploadId>"
+                    b"</InitiateMultipartUploadResult>"
+                )
+                self._send(200, body, close=False)
+                return
+            self._send(404, b"<Error/>", close=False)
+
+        def do_PUT(self):
             body = (
                 b'<?xml version="1.0" encoding="UTF-8"?>'
-                b"<InitiateMultipartUploadResult>"
-                b"<Bucket>fake</Bucket><Key>fake</Key>"
-                b"<UploadId>fake-upload-id</UploadId>"
-                b"</InitiateMultipartUploadResult>"
+                b"<Error><Code>XMinioStorageFull</Code>"
+                b"<Message>Storage backend has reached its capacity</Message></Error>"
             )
-            self._send(200, body)
-            return
-        self._send(404, b"<Error/>")
+            content_len = int(self.headers.get("Content-Length", "0"))
+            if content_len:
+                self.rfile.read(content_len)
+            self._send(507, body, close=connection_close)
 
-    def do_PUT(self):
-        body = (
-            b'<?xml version="1.0" encoding="UTF-8"?>'
-            b"<Error><Code>XMinioStorageFull</Code>"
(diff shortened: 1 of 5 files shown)


util/fibers/fiber_socket_base.cc
```diff
@@ -181,11 +181,13 @@ error_code LinuxSocketBase::Listen(unsigned backlog) {
 }
 
 auto LinuxSocketBase::Shutdown(int how) -> error_code {
-  CHECK_GE(fd_, 0);
+  DCHECK_GE(fd_, 0);
```
glevkovich, line 184: I think this should be removed. if you don't remove, we will crash again in DEBUG tests.
Consider replacing this line with 
`DLOG(WARNING) << "Shutdown called with invalid fd: " << fd_;`
Or just completely remove it.

util/cloud/utils.cc
```diff
@@ -235,6 +238,7 @@ error_code RobustSender::Send(unsigned num_iterations, detail::HttpRequestBase*
       VLOG(1) << "Refreshing token";
       RETURN_ERROR(provider_->RefreshToken());
       provider_->Sign(req);
+      ec = make_error_code(errc::permission_denied);
```
glevkovich, line 241: nit: 
I see there is no documentation  for the function in  the header.

So we have here permission_denied set as error , and then we continue the loop. If this happens to be the final iteration of the loop, the function will exit and return permission_denied to the caller. This masks the fact that the actual failure was retry exhaustion, not just a permission error.

I think we should add a brief inline comment near the final return ec or in the header: explaining that ec reflects the error of the last retry attempt, which may just be the trigger for the retry exhaustion. It is not trivial.

Example:

Consider this sequence across 3 iterations:
Iteration 0: server returns 507              ->  ec = resource_unavailable_try_again, retry
Iteration 1: server returns 507 again   ->  same, retry
Iteration 2 (last): server returns 401    ->  token refreshed, ec = permission_denied, loop ends

ode is right, but caller need to better understand the returned value.

tests/test_s3.py
```diff
@@ -169,6 +172,163 @@ def test_endpoint_flag_override(minio):
...
+        "AWS_ACCESS_KEY_ID": "fake",
+        "AWS_SECRET_ACCESS_KEY": "fake",
+        "AWS_DEFAULT_REGION": "us-east-1",
+    })
+    return _run(
+        binary,
+        "--cmd=put-object",
+        "--bucket=fake-bucket",
+        "--key=fake/key.bin",
+        "--upload_size=16",
```
glevkovich: nit:
 `--upload_size=16` - why is it 16B only? - I think it worth a one-line comment explaining the small size:

```
"--upload_size=16",  # s3_storage always uses multipart regardless of size. keep small for speed.
```

I can see that The helio S3 implementation (`OpenWriteFile`) unconditionally initiates a multipart upload ("// Initiate multipart upload.") - there is no size threshold check. In production this is fine since snapshots are always large, but it also means 16 bytes exercises the exact same `CreateMultipartUpload` -> `UploadPart` code path as a real backup would.

util/fibers/fiber_socket_base.cc
romange: why will crash? we now check on the caller side whether the socket is open. I consider operating on sockets with closed state as a bug and it's consistent with other methods.

util/cloud/utils.cc
romange: fixed

tests/test_s3.py
romange: fixed

util/fibers/fiber_socket_base.cc
glevkovich: OK understood, you want to crash on debug. So drop this comment.

glevkovich (approved):
LGTM!

PR #592, opened 2026-05-26
chore: starvation of background fibers under high task queue load

kostasrim:
When a proactor's task queue is heavy, it sets `has_cpu_work` to true. Under high and continuous load the `background` fiber queue is `starved`. The root cause is that the proactor run loop skips the BACKGROUND tasks because has_cpu_work is true. This spins without a fallback for the entire load time leading to the starvation of the BACKGROUND task queue. 

CI being red is the proof the tests fails. I will follow up with a fix.

### DIFF
--- a/util/fibers/fibers_test.cc
+++ b/util/fibers/fibers_test.cc
@@ -1086,6 +1086,62 @@ TEST_P(ProactorTest, Background) {
   fb.Join();
 }
 
+// Regression test for background fiber starvation.
+//
+// Starvation path:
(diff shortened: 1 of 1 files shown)


kostasrim:
@romange irrelevant to the heartbeat issue I am investigating I found this.  iouring proactor run loop can starve background fibers when load on the task queue is high

I don't see this being critical for now but it's worth fixing (:

romange:
@dranikpg can comment on this as he designed this feature.

PR #593, opened 2026-05-27
chore: tls socket NEED_WRITE on recv while WRITE_IN_PROGRESS

kostasrim:
We had a test in dragonfly that used protocol renegotiation to exercise the same path. However it was unreliable and reproduced very rarely. 1.3 connections use KeyUpdates to exercise the same path (in fact it's the only case the Recv might NEED_WRITE).

I will also remove the test from dragonfly.

* remove DFATAL log
* update comment
* add test
* small refactor -- use absl::Cleanup in tls_socket_test

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents a fatal failure and potential deadlock during concurrent TLS write/flush by waiting and proceeding safely instead of aborting.
  * Ensures pending TLS output is flushed correctly after edge cases like connection reset or TLS key-update.
  * Reduces hard failures on accept failures by logging and returning early instead of crashing.

* **Tests**
  * Added a Linux-only TLS 1.3 regression/stress test exercising concurrent read/write and key-update races with non-blocking sockets.
  * Improved test resource cleanup using scoped guards to reduce flakiness.

* **Chores**
  * Increased CI test timeout to allow longer test runs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/romange/helio/pull/593?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/util/tls/tls_socket.cc
+++ b/util/tls/tls_socket.cc
@@ -392,11 +392,12 @@ auto TlsSocket::MaybeSendOutput() -> error_code {
 
 auto TlsSocket::HandleUpstreamRead() -> error_code {
   if (engine_->OutputPending() != 0) {
-    // All tls operations should make sure that output is flushed before finisihing.
-    // If the state machine requested reading, then the output has been flushed,
-    // or there is a concurrent write in progress.
+    // Normally output is flushed before the read path needs upstream data, or a concurrent
+    // write fiber is already in progress (WRITE_IN_PROGRESS). During a TLS 1.3 KeyUpdate the
+    // write fiber may exit (e.g. after a connection reset) without draining the ack, leaving
+    // OutputPending > 0 with WRITE_IN_PROGRESS clear. Flush here; MaybeSendOutput will either
+    // succeed or return the connection error, which is the right outcome either way.
     if ((state_ & WRITE_IN_PROGRESS) == 0) {
-      LOG(DFATAL) << "Should not happen";
       RETURN_ON_ERROR(MaybeSendOutput());
     }
   }
--- a/util/tls/tls_socket.cc
+++ b/util/tls/tls_socket.cc
@@ -370,24 +370,17 @@ auto TlsSocket::MaybeSendOutput() -> error_code {
   if (engine_->OutputPending() == 0)
     return {};
 
-  // This function is present in both read and write paths.
-  // meaning that both of them can be called concurrently from differrent fibers and then
-  // race over flushing the output buffer. We use state_ to prevent that.
+  // Called from both the read and write paths, which may run concurrently on different fibers.
+  // On the write path this is straightforward. On the read path, the TLS engine can generate
+  // outbound data that must be flushed before reading can continue — this happens during
+  // protocol renegotiation (TLS 1.2) or a KeyUpdate exchange (TLS 1.3), both of which cause
+  // SSL_read to return SSL_ERROR_WANT_WRITE.
+  // In that case a concurrent write fiber may already hold WRITE_IN_PROGRESS. We must wait for
+  // it to finish rather than spinning: without yielding, the fiber spins indefinitely and never
+  // passes control to anyone else, so WRITE_IN_PROGRESS never clears and the engine loops on
+  // NEED_WRITE forever.
+  // See Tls13KeyUpdateNeedWrite test for the concrete scenario.
   if (state_ & WRITE_IN_PROGRESS) {
-    // This should not happen, as we call MaybeSendOutput from reads only when
-    // there is no ongoing write.
-    LOG(DFATAL) << "Should not happen";
-
-    // Fiber1 (Connection fiber):
-    // Reads SSL renegotiation request
-    // Sets WRITE_IN_PROGRESS in state_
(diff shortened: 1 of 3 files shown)


util/tls/tls_socket.cc
```diff
@@ -370,24 +370,17 @@ auto TlsSocket::MaybeSendOutput() -> error_code {
...
-    // Reads SSL renegotiation request
-    // Sets WRITE_IN_PROGRESS in state_
-    // Tries to write renegotiation response to socket
-    // BIO (SSL buffer) gets filled up
-    // Fiber2 (runs command):
-    // Tries to write response to socket
-    // Can't write to socket because WRITE_IN_PROGRESS is set.
-
-    // Wait for the other write to complete.
     fb2::NoOpLock lock;
```
kostasrim, line 384: We can not remove this lock. Yes the WRITE fiber will eventually flush the engine to the socket but the READ fiber *must* yield so the WRITE fiber can get a progress guarantee. Otherwise, the READER will loop indefinitely and will deadlock (engine -> needs write -> early return -> loop).

I updated the comment

util/tls/tls_socket_test.cc
```diff
@@ -244,10 +249,114 @@ TEST_P(TlsSocketTest, KtlsClientFdApiFeasibility) {
...
+  // proceed.
+  //
+  // Sequence after close(fd) → TCP RST:
+  //   1. Write fiber's WriteSome fails (ECONNRESET), exits without draining the KeyUpdate ack,
+  //      clears WRITE_IN_PROGRESS, notify_one().
+  //   2. Read fiber wakes from MaybeSendOutput's cv.wait, returns {}.
+  //   3. RecvMsg loops, engine_->Read() returns NEED_READ_AND_MAYBE_WRITE — ack still in output
+  //      buffer because the write fiber exited before draining it.
+  //   4. HandleUpstreamRead sees OutputPending() > 0 with WRITE_IN_PROGRESS clear, calls
+  //      MaybeSendOutput() which fails with ECONNRESET; error propagates and read fiber exits.
```
kostasrim, line 357: @romange two birds one stone. see also https://github.com/romange/helio/pull/593/changes#diff-e23cdb5eee2e5d12ad3b9853e48dcfdc99ddbb3670ee960206c491dbcbe79c5aR400

util/tls/tls_socket_test.cc
```diff
@@ -244,10 +252,132 @@ TEST_P(TlsSocketTest, KtlsClientFdApiFeasibility) {
...
+  LOG(INFO) << "Closing client connection";
+  { auto _ = std::move(ssl_guard); }  // SSL_shutdown + SSL_free (non-blocking, may be EAGAIN)
+  LOG(INFO) << "ssl_guard destroyed, closing fd";
+  { auto _ = std::move(fd_guard); }   // close() → RST due to unread data
+  LOG(INFO) << "fd_guard destroyed (RST sent), waiting for server fibers";
+
+  // Watchdog: if the server fibers don't unblock within 10s, dump all fiber stacks and abort.
+  fb2::Done watchdog_done;
```
kostasrim, line 365: So it took me some while to figure out. Sometimes, a test run can take close to 2 minutes. See for example: https://github.com/romange/helio/actions/runs/26574428238/job/78289768895 

And my PR failed with CI timing out and I thought it was a deadlock. It is not because https://github.com/romange/helio/actions/runs/26586351000 if you compare the timestamps, we had 1 second left to finish the test execution which is not enough. 

I increased the CI timeout time to 3minutes (from 2) and for safety reasons, I added a watchdog just in case we encounter a timeout again.

P.s. I run the test a thousand times without any failure

util/tls/tls_socket.cc
```diff
@@ -399,11 +392,12 @@ auto TlsSocket::MaybeSendOutput() -> error_code {
 
 auto TlsSocket::HandleUpstreamRead() -> error_code {
   if (engine_->OutputPending() != 0) {
-    // All tls operations should make sure that output is flushed before finisihing.
-    // If the state machine requested reading, then the output has been flushed,
-    // or there is a concurrent write in progress.
+    // Normally output is flushed before the read path needs upstream data, or a concurrent
+    // write fiber is already in progress (WRITE_IN_PROGRESS). During a TLS 1.3 KeyUpdate the
```
romange, line 396: so we exit due to error during socket write so concurrent read can reach this state?

romange:
@coderabbitai review

romange:
@coderabbitai help

romange:
@coderabbitai full review

romange (approved):

util/tls/tls_socket.cc
kostasrim: yes 

server side write blocks because client is not reading
server side read gets a NEED_WRITE blocks on cv because WRITE_IN_PROGRESS (1)
client closes the socket
server side write receives a conn reset and exits
server side read wakes up (cv notified), engine NEED_READ_AND_MAYBE_WRITE (2) -> MaybeSendOutput -> fails because socket is closed and we return an error

The numbers in parenthesis (1) and (2) are the DCHECK triggered at each step

PR #594, opened 2026-06-02
feat: add build dir param into blaze.sh

BorysTheDev:
Added parameter `-build-dir arg` into blaze.sh to choose what a directory will be used for building

## Summary by CodeRabbit

* **New Features**
  * Added a -build-dir <path> option to specify a custom build directory (validated; falls back to the previous default).
  * Now recognizes and forwards user-supplied CMake-style -D... options to the build invocation.
  * When a compiler launcher (e.g., ccache) is available, it is automatically used for C++ builds.

* **Behavior Change**
  * The set of CLI flags forwarded to the build step has been tightened to include only recognized CMake options.

### DIFF
--- a/blaze.sh
+++ b/blaze.sh
@@ -4,16 +4,20 @@ TARGET_BUILD_TYPE=Debug
 BUILD_PREF=build
 BUILD_SUF=dbg
 GENERATOR='-GNinja'
-LAUNCHER=$(command -v ccache)
-# coloring
+
 NC=$'\e[0m'
 RED=$'\e[0;31m'
 GREEN=$'\e[0;32m'
-if [ -x "$LAUNCHER" ]; then
-  echo "Using launcher $LAUNCHER"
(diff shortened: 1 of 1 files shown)


blaze.sh
```diff
@@ -4,16 +4,20 @@ TARGET_BUILD_TYPE=Debug
 BUILD_PREF=build
 BUILD_SUF=dbg
 GENERATOR='-GNinja'
-LAUNCHER=$(command -v ccache)
```
romange, line 7: why `LAUNCHER` is removed?

romange (approved):

blaze.sh
BorysTheDev: it is not removed

BorysTheDev: it had small bug, so it was fixed with CMAKE_EXTRA+=("-DCMAKE_CXX_COMPILER_LAUNCHER=$CCACHE_BIN")

PR #600, opened 2026-06-14
mpmc_bounded_queue: add fetch-and-add blocking enqueue for FiberQueue

romange:
## Summary

Adds a fetch-and-add (FAA) enqueue discipline to `mpmc_bounded_queue` for callers that are allowed to block, and rewrites `FiberQueue::Add` to use it.

- **`claim_slot` / `slot_ready` / `commit_slot`** — a blocking caller claims a ticket with a single `fetch_add` (never fails, no retry), waits until its cell is free, then commits. This avoids the cache-line contention of the `try_enqueue` CAS retry loop on `enqueue_pos_` under many producers.
- The per-cell sequence protocol is shared with `try_enqueue`, so the FAA and CAS disciplines coexist on the same queue.
- `FiberQueue::Add` now claims a slot, yields once, then blocks on `push_ec_` until the slot frees; the `blocked_submitters_` contention gauge moves into `FiberQueue`.
- `try_enqueue` / `TryAdd` are kept for the non-blocking path (`AddAnyWorker`), which still needs "detect full and bail" semantics FAA can't provide.
- Adds unit tests, a CAS-vs-FAA micro-benchmark, and a `FiberQueue` blocking-path test.

## Benchmark

`BM_MPMCContention` vs `BM_MPMCContentionFAA`, 8 producers / 1 consumer, q=1024, 2²⁰ ops/producer, identical workload:

| discipline | throughput | per-op counter |
|---|---|---|
| CAS (`try_enqueue`) | **11.4 M items/s** | enq_fail/op = 734µ |
| FAA (`claim_slot`+`commit_slot`) | **32.4 M items/s** | enq_spin/op = 0.45 |

FAA is **~2.8× faster**. Key finding: `enq_fail/op` is ~0.0007, so the queue is almost never full — the CAS path's cost is **not** full-queue backoff but the `compare_exchange_weak` loop on the shared `enqueue_pos_` cache line. Lost CAS races re-loop invisibly (the counter only ticks on a full queue), showing up purely as wall-clock. FAA replaces that with one unconditional `fetch_add`, so every producer makes progress on the first try. The only remaining wait is enq_spin/op = 0.45, a spin on `slot_ready()` that loads the producer's own (uncontended) cell while the consumer drains it.

## Algorithm lineage & prior art

This is not a new algorithm — it is a composition of two well-known designs, specialized for a fiber runtime. Credit where it is due:

- **Dmitry Vyukov — bounded MPMC queue** ([1024cores](https://www.1024cores.net/home/lock-free-algorithms/queues/bounded-mpmc-queue)). `mpmc_bounded_queue` already implements his per-cell `sequence` protocol (cell `i` starts at `sequence = i`; publish bumps it to `pos + 1`; dequeue frees it to `pos + mask + 1`). This PR reuses that protocol verbatim — `slot_ready` / `commit_slot` are just the existing publish steps with the CAS removed.
- **Erik Rigtorp — `MPMCQueue`** ([github](https://github.com/rigtorp/MPMCQueue), [writeup](https://rigtorp.se/ringbuffer/)). The "claim a slot with a single `fetch_add`, then wait on that slot's turn before writing" idea is exactly Rigtorp's `push`: `auto head = head_.fetch_add(1); … while (turn(head) != slot.turn) …`. Rigtorp's own code also validates the CAS-vs-FAA split used here: his **blocking** `push`/`pop` use `fetch_add`, while his **non-blocking** `try_push`/`try_pop` use `compare_exchange` — the same reason we keep `try_enqueue` on CAS for the non-blocking `AddAnyWorker` path.
- **LCRQ — Morrison & Afek, PPoPP 2013** (["Fast Concurrent Queues for x86 Processors"](https://www.cs.tau.ac.il/~mad/publications/ppopp2013-x86queues.pdf)). The academic root of using `fetch_add` to claim a cell as the way to beat CAS-loop contention.

### How this differs from Rigtorp's implementation

- **Blocking, not spinning.** Rigtorp's blocking `push`/`pop` busy-spin on the slot turn (`while (…) {}`, no sleep). On a fiber runtime that pegs a core and stalls the whole proactor thread. Here the primitives are fiber-agnostic (pure atomics, no `EventCount`), and `FiberQueue::Add` owns the waiting — it yields once, then suspends on `push_ec_` — so the proactor keeps running other fibers while a producer waits on a full queue.
- **Single-consumer.** Rigtorp's `pop` does `tail_.fetch_add` (multi-consumer). We pair the FAA enqueue with the existing single-consumer `try_dequeue_sc` and a batch drain that wakes blocked producers, with no consumer-side atomic arbitration.
- **No vendored dependency.** Because Vyukov's `sequence` already provides the per-slot "turn", we extend the queue we already ship rather than pulling in a third-party header. Rigtorp is the reference; Vyukov's queue is the implementation.

## Release Notes

* **New Features**
  * Added a slot-based enqueue workflow to bounded MPMC queues, including inflight-slot visibility and readiness checks.
  * Exposed a per-thread indicator for blocked submitters in the fiber queue threadpool.
* **Refactor**
  * Updated the fiber queue threadpool enqueue path to use the slot-based workflow and prevent premature worker shutdown while enqueues are in progress.
  * Refined internal task representation to a fixed-capacity callable.
* **Tests**
  * Added bounded MPMC queue tests for ordering, wrap-around behavior, inflight semantics, move-only correctness, and mixed producer stress.
  * Added fiber queue threadpool tests covering blocking behavior and await result handling.
* **Benchmarks**
  * Extended MPMC contention benchmarks with enqueue-failure metrics and added a dedicated FAA-focused contention benchmark.

### DIFF
--- a/base/mpmc_bounded_queue.h
+++ b/base/mpmc_bounded_queue.h
@@ -135,8 +135,12 @@ template <typename T> class mpmc_bounded_queue {
 
   // Unconditionally claims a ticket. Never fails. The caller MUST eventually commit_slot(pos),
   // waiting for slot_ready(pos) first if it is not already free.
+  //
+  // relaxed is sufficient: enqueue_pos_ only hands out unique tickets, it carries no data. All
+  // data publication is done by commit_slot's release store and observed by slot_ready's /
+  // try_dequeue's acquire load. This matches try_enqueue, which also bumps enqueue_pos_ relaxed.
   size_t claim_slot() {
-    return enqueue_pos_.fetch_add(1, std::memory_order_acq_rel);
+    return enqueue_pos_.fetch_add(1, std::memory_order_relaxed);
   }
 
   // Returns true once the cell for ticket pos is free to be written (its previous occupant, if
@@ -153,6 +157,15 @@ template <typename T> class mpmc_bounded_queue {
     cell_t& cell = buffer_[pos & buffer_mask_];
     new (&cell.storage) T(std::forward<U>(data));
     cell.sequence.store(pos + 1, std::memory_order_release);
+  }
+
+  // True if any ticket has been claimed (claim_slot or try_enqueue) but not yet dequeued. A
+  // blocking caller that has claimed a slot but not yet committed it shows up here as a gap that
+  // try_dequeue cannot see, so a single-consumer shutdown can use this to avoid exiting while a
+  // commit_slot is still outstanding. Cache coherence makes the claimed enqueue_pos_ bump visible
+  // here promptly, so relaxed loads suffice for this coordination check.
+  bool has_inflight_slots() const {
+    return enqueue_pos_.load(std::memory_order_relaxed) != dequeue_pos_.load(std::memory_order_relaxed);
   }
 
   bool try_dequeue(T& data) {
--- a/base/mpmc_bounded_queue.h
+++ b/base/mpmc_bounded_queue.h
@@ -107,6 +107,54 @@ template <typename T> class mpmc_bounded_queue {
     return true;
   }
 
+  // Fetch-and-add enqueue discipline (claim_slot / slot_ready / commit_slot).
+  //
+  // The "claim a slot with fetch_add, then wait on that slot's turn" idea follows Erik Rigtorp's
+  // MPMCQueue (https://github.com/rigtorp/MPMCQueue, https://rigtorp.se/ringbuffer/), which is in
+  // the lineage of LCRQ (Morrison & Afek, PPoPP 2013). Rigtorp's blocking push/pop busy-spin on the
+  // slot turn and the queue is multi-consumer; here we keep Vyukov's per-cell sequence (above) as
+  // the "turn", expose only fiber-agnostic primitives so the caller can block instead of spin, and
+  // pair them with the single-consumer try_dequeue_sc.
+  //
+  // try_enqueue uses a CAS retry loop so it can detect a full queue and back off *without*
+  // committing. A caller that is allowed to block does not need that: it can claim a slot
+  // unconditionally with a single fetch_add (no CAS, no retry, never fails), then wait until the
+  // slot is free, then commit. This avoids the cache-line contention of the CAS loop on
+  // enqueue_pos_ under many producers.
+  //
+  // The per-cell sequence protocol is shared with try_enqueue, so the two disciplines may be mixed
+  // on the same queue: both advance enqueue_pos_ by exactly one, keeping tickets dense and unique,
+  // and each producer serializes on its own cell via the sequence value.
+  //
+  // These primitives are intentionally fiber-agnostic (pure atomics, no blocking). The blocking
+  // (e.g. via EventCount) is the caller's responsibility and must happen between claim_slot() and
+  // commit_slot(). Typical usage:
+  //   size_t pos = q.claim_slot();
+  //   while (!q.slot_ready(pos)) { /* caller blocks/yields */ }
+  //   q.commit_slot(pos, std::move(item));
+
+  // Unconditionally claims a ticket. Never fails. The caller MUST eventually commit_slot(pos),
+  // waiting for slot_ready(pos) first if it is not already free.
+  size_t claim_slot() {
+    return enqueue_pos_.fetch_add(1, std::memory_order_acq_rel);
+  }
+
+  // Returns true once the cell for ticket pos is free to be written (its previous occupant, if
+  // any, has been fully dequeued). The acquire load pairs with the dequeue's release store.
+  bool slot_ready(size_t pos) const {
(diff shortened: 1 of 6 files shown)


util/fibers/fiberqueue_threadpool.h
```diff
@@ -36,26 +36,39 @@ class FiberQueue {
...
+      --blocked_submitters_;
     }
 
-    bool result = false;
-    while (true) {
-      auto key = push_ec_.prepareWait();
+    queue_.commit_slot(pos, std::forward<F>(f));
+    pull_ec_.notify();
+    return preempted;
+  }
```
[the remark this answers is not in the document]
romange: fixed

base/mpmc_bounded_queue.h
```diff
@@ -107,6 +107,47 @@ template <typename T> class mpmc_bounded_queue {
...
+  // commit_slot(). Typical usage:
+  //   size_t pos = q.claim_slot();
+  //   while (!q.slot_ready(pos)) { /* caller blocks/yields */ }
+  //   q.commit_slot(pos, std::move(item));
+
+  // Unconditionally claims a ticket. Never fails. The caller MUST eventually commit_slot(pos),
+  // waiting for slot_ready(pos) first if it is not already free.
+  size_t claim_slot() {
+    return enqueue_pos_.fetch_add(1, std::memory_order_acq_rel);
+  }
```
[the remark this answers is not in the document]
romange: fixed

util/fibers/fiberqueue_threadpool.h
```diff
@@ -36,26 +36,39 @@ class FiberQueue {
...
+
+    bool preempted = false;
+    if (!queue_.slot_ready(pos)) {
+      preempted = true;
+      ++blocked_submitters_;
+      detail::FiberActive()->Yield();
+      if (!queue_.slot_ready(pos)) {
+        push_ec_.await([&] { return queue_.slot_ready(pos); });
+      }
+      --blocked_submitters_;
```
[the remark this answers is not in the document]
romange: not relevant

util/fibers/fiberqueue_threadpool.h
```diff
@@ -36,26 +54,39 @@ class FiberQueue {
...
     }
 
-    bool result = false;
-    while (true) {
-      auto key = push_ec_.prepareWait();
+    queue_.commit_slot(pos, std::forward<F>(f));
+    pull_ec_.notify();
```
dranikpg, line 82: the CAS was on enqueue_pos - now it is predetermined. However wouldn't having slot_ready + commit as a single CAS not benefit?

dranikpg: Also the yield here is quite an interesting choice. First, it's quite a heavy time skip under load, second - with inter-thread coordination the timing is off either way

dranikpg (dismissed):

util/fibers/fiberqueue_threadpool.h
dranikpg: also what I don't understand: if we now pre-assign the pos (like with a ticket system) and Yield, then we will stall the whole qeue, won't we? (I know same can happen is OS takes thread off in any atomic data structure, but this is one more point here)

romange: 1. CAS is heavier as it is a full blown transaction. Factually, you can look at the unit test and run it yourself locally and see yourself: it compares FAA and CAS. In addition it is not my invention. FAA class of algorithms are known to be more efficient that CAS due to their "blind" semantics. I am sure gemini can explain better than me why CAS is heavier across cpus.
2. Regarding Yield - i am not sure . We can try a simple (thread) spinning too but my guess is that won't help. You need to understand  the use-case of when a slot can be busy. It's not about producer contention. For example, for a queue say at capacity 16, suppose we claim cell 0 at `pos` 32 and its busy. It necessary means that the queue is full. it may be that tail have not even started consuming at all, i.e. cell 0 has pos 0.
but maybe it consumed cell 0 with pos 0, but then a producer claimed that cell with pos 16 but consumer has not consumed it (it is somewhere between cells [1, 15].  We may have even 100 producers waiting on the same queue cell 0 but with unique `pos` slots : 32, 48, 64 etc. They will wait for  a) consumer dequeuing from cell 0  and b) their turn (pos) to enqueue.  So `claim`/`ready`/`commit` is not only for atomics contention - it is a transactional algorithm that also ensures fairness and order.  
5. We do not stall the queue. Again, say we enqueue at `pos` (our head) and it's busy it means the queue is full, we do not think we can claim a cell that is busy in a non-full queue. 

At least this is how I understand the algorithm. What I really like about this is that we can implement these differrent semantics using the same data-structure. It is very elegant, imho

PR #603, opened 2026-07-02
chore(proactor): organize statistics around completions gathering

romange:
Also, remove wrong optimization where we break from the completion loop early. I verified it does not hold and with small batch we can still get next one if we peek.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved completion-event reaping and CQ draining across proactors to better process queued work and lower stall risk.
  * Updated completion counting so reported metrics reflect the actual number of processed completions.

* **Improvements / Observability**
  * Refined shutdown and end-of-loop diagnostics to reduce log noise and highlight the most relevant counters.
  * Enhanced per-run proactor logging with additional run statistics and a conditional completion ratio.
  * Adjusted submission/counter timing and idle pacing to improve metric consistency.

### DIFF
--- a/util/fibers/epoll_proactor.cc
+++ b/util/fibers/epoll_proactor.cc
@@ -145,7 +145,7 @@ EpollProactor::~EpollProactor() {
   CHECK(is_stopped_);
   close(epoll_fd_);
 
-  DVLOG(1) << "~EpollProactor";
+  DVLOG(2) << "~EpollProactor";
 }
 
 void EpollProactor::Init(unsigned pool_index) {
@@ -284,12 +284,10 @@ void EpollProactor::MainLoop(detail::Scheduler* scheduler) {
       while (true) {
         VPRO(2) << "Fetched " << cqe_count << " cqes";
         DispatchCompletions(&ev_batch, cqe_count);
+        stats_.num_completions += cqe_count;
 
-        if (cqe_count < kEvBatchSize) {
-          break;
-        }
         epoll_res = EpollWait(epoll_fd_, &ev_batch, 0);
-        if (epoll_res < 0) {
+        if (epoll_res <= 0) {
           break;
         }
         cqe_count = epoll_res;
@@ -317,9 +315,6 @@ void EpollProactor::MainLoop(detail::Scheduler* scheduler) {
       scheduler->DestroyTerminated();
     }
   }
-
-  VPRO(1) << "total/stalls/cqe_fetches/num_suspends: " << stats_.loop_cnt << "/"
-          << stats_.num_stalls << "/" << stats_.completions_fetches << "/" << stats_.num_suspends;
 
   VPRO(1) << "wakeups/stalls/task_int: " << tq_wakeup_ev_.load() << "/" << stats_.task_interrupts;
   VPRO(1) << "centries size: " << centries_.size();
--- a/util/fibers/proactor_base.cc
+++ b/util/fibers/proactor_base.cc
@@ -99,12 +99,11 @@ std::once_flag module_init;
 __thread ProactorBase::TLInfo ProactorBase::tl_info_;
 
 ProactorBase::Stats& ProactorBase::Stats::operator+=(const Stats& other) {
-  static_assert(sizeof(Stats) == 9 * sizeof(uint64_t));
+  static_assert(sizeof(Stats) == 8 * sizeof(uint64_t));
 
   num_stalls += other.num_stalls;
   completions_fetches += other.completions_fetches;
   loop_cnt += other.loop_cnt;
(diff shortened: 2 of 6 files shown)


util/fibers/uring_proactor.cc
```diff
@@ -346,15 +346,17 @@ void UringProactor::ProcessCqeBatch(unsigned count, io_uring_cqe** cqes,
...
-    reaped_cqe_cnt_ += batch_count;
-    if (batch_count < kCqeBatchLen)
-      break;
+    stats_.num_completions += batch_count;
+
+    // There could be completions added during the processing of the batch,
+    // so we check again.
     batch_count = io_uring_peek_batch_cqe(&ring_, cqes, kCqeBatchLen);
-  }
+  } while (batch_count > 0);
```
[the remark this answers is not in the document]
romange: irrelevant - number of completions is bounded and we want to reap larger batches.

util/fibers/proactor_base.cc
```diff
@@ -410,6 +426,15 @@ void ProactorBase::ModuleInit() {
 void ProactorDispatcher::Run(detail::Scheduler* sched) {
   proactor_->cpu_measure_cycle_start_ = base::CycleClock::Now();
   proactor_->MainLoop(sched);
+  auto& stats = proactor_->stats();
+
+  VLOG(1) << "PRO[" << proactor_->GetPoolIndex()
+          << "] total_loop/stalls/cqe_fetches/num_submits: " << stats.loop_cnt << "/"
+          << stats.num_stalls << "/" << stats.completions_fetches << "/"
+          << stats.uring_submit_calls;
```
glevkovich: nit: uring_submit_calls belongs to a, UringProactor so for epoll this will print 0 always, and might confuse reader. Consider creating a pure virtual function that each  proactor must override `std::string StatsSummary() ` - and return proator's specific stats string.

util/fibers/uring_proactor.cc
```diff
@@ -954,7 +954,7 @@ void UringProactor::MainLoop(detail::Scheduler* scheduler) {
     // and we enter too often into kernel space.
     bool should_poll = GetCurrentBusyCycles() < busy_poll_cycle_limit_;
     if (!ring_busy && should_poll) {
-      Pause(3);
+      Pause(2);
```
glevkovich: nit: unrelated to PR. 1) Consider in different Pr or 2) At least - consider having it in a separate commit and merge the PR as "Rebase and Merge".

util/fibers/proactor_base.h
```diff
@@ -197,8 +197,12 @@ class ProactorBase {
   struct Stats {
     uint64_t num_stalls = 0, completions_fetches = 0, loop_cnt = 0, num_suspends = 0;
```
glevkovich: num_suspends is a dead code.
remove it.

util/fibers/proactor_base.cc
romange: no, I need to return stats object. The user won't be confused :)

util/fibers/proactor_base.cc
```diff
@@ -410,6 +425,15 @@ void ProactorBase::ModuleInit() {
...
```
glevkovich: How about printing here also `uring_submit_sqes` in the print in line 430, and also add after the block on line 434 the similar relationship:

```
if (stats.uring_submit_calls > 0)
  VLOG(1) << "PRO[" << proactor_->GetPoolIndex()
          << "] AvgSqe/SubmitCall: " << double(stats.uring_submit_sqes) / stats.uring_submit_calls;
```

util/fibers/epoll_proactor.cc
```diff
@@ -284,12 +284,10 @@ void EpollProactor::MainLoop(detail::Scheduler* scheduler) {
       while (true) {
         VPRO(2) << "Fetched " << cqe_count << " cqes";
         DispatchCompletions(&ev_batch, cqe_count);
+        stats_.num_completions += cqe_count;
 
-        if (cqe_count < kEvBatchSize) {
```
glevkovich, line 288: nit: I would move this till line 294 in old code too into a separate commit because this is a behaviour change, not related ot stats.

glevkovich (dismissed):
approved, there are few minor comments.

util/fibers/epoll_proactor.cc
romange: it's a bug fix, the PR description mentions it

util/fibers/proactor_base.cc
romange: I will remove ratio printings as metrics allow me to see it in grafana.

glevkovich (approved):

PR #605, opened 2026-07-05
chore: Improvements around FiberQueue

romange:
## Summary
Replaces `FiberQueue`'s callback storage with a fixed-capacity `fu2::function` (`Tasklet`), separates
the producer/consumer `EventCount`s onto distinct cache lines, and adds a `blocked_submitters_` metric.

## Changes
- Replace `std::function<void()>` (`CbFunc`) with `Tasklet`, a fixed-capacity (32 bytes inline, 8-byte
  alignment), non-copyable, non-throwing `fu2::function_base`. Callbacks up to 32 bytes avoid heap
  allocation. Ported from #595.
- `alignas(64)` `pull_ec_` and `push_ec_` separately — previously adjacent, they shared a cache line
  and ping-ponged between producer-side (`notify()` per commit) and consumer-side (`notifyAll()` per
  drained batch) wakeup traffic.
- Add a `blocked_submitters_` thread-local counter, incremented/decremented around
  `push_ec_.wait()` when a submitter blocks on a full queue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task submission and execution behavior under heavy load and short queue-full conditions by streamlining batch handling to avoid unnecessary exception overhead.
  * Strengthened backpressure responsiveness via per-thread tracking of blocked submitters.
* **Refactor / Performance**
  * Updated the internal task execution pipeline to use a fixed-capacity callable wrapper and tasklet-based queueing for more consistent throughput.
* **New Features**
  * Added a new `blocked_submitters()` metric to monitor per-thread queue contention levels.

### DIFF
--- a/util/fibers/fiberqueue_threadpool.cc
+++ b/util/fibers/fiberqueue_threadpool.cc
@@ -14,6 +14,8 @@ namespace fb2 {
 
 using namespace std;
 
+__thread unsigned FiberQueue::blocked_submitters_ = 0;
+
 FiberQueue::FiberQueue(unsigned queue_size) : queue_(queue_size) {
 }
 
@@ -21,7 +23,7 @@ void FiberQueue::Run() {
   static constexpr unsigned kBatchSize = 16;
 
   bool is_closed = false;
-  CbFunc batch[kBatchSize];
+  Tasklet batch[kBatchSize];
   unsigned count = 0;
 
   auto cb = [&] {
@@ -49,11 +51,7 @@ void FiberQueue::Run() {
     push_ec_.notifyAll();
 
     for (unsigned i = 0; i < count; ++i) {
-      try {
-        batch[i]();
-      } catch (std::exception& e) {
-        LOG(FATAL) << "Exception " << e.what();
-      }
+      batch[i]();
     }
     count = 0;
   }
--- a/util/fibers/fiberqueue_threadpool.h
+++ b/util/fibers/fiberqueue_threadpool.h
@@ -7,6 +7,11 @@
 #include "util/fibers/detail/result_mover.h"
 #include "util/fibers/synchronization.h"
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Waddress"
+#include "base/function2.hpp"
+#pragma GCC diagnostic pop
+
 namespace util {
 namespace fb2 {
 class FiberQueueThreadPool;
@@ -18,6 +23,19 @@ class FiberQueueThreadPool;
  */
 class FiberQueue {
   friend class FiberQueueThreadPool;
+
+  // Fixed-capacity, non-copyable, non-throwing callable. Callbacks up to 32 bytes are stored
+  // inline (no heap allocation), unlike std::function. Keeps queue cells trivially relocatable.
+  using Fu2Fun =
+      fu2::function_base<true /* owns */, false /*non-copyable*/, fu2::capacity_fixed<32, 8>,
+                         false /* non-throwing*/, false /* strong exceptions guarantees*/, void()>;
+
+  struct Tasklet : public Fu2Fun {
+    using Fu2Fun::Fu2Fun;
+    using Fu2Fun::operator=;
+  };
+
+  static_assert(sizeof(Tasklet) == 48);
 
  public:
   explicit FiberQueue(unsigned queue_size = 128);
@@ -52,8 +70,11 @@ class FiberQueue {
       if (TryAdd(std::forward<F>(f))) {
         break;
       }
+
       result = true;
+      blocked_submitters_++;
       push_ec_.wait(key.epoch());
+      blocked_submitters_--;
     }
     return result;
   }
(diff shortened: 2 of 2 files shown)


util/fibers/fiberqueue_threadpool.cc
```diff
@@ -49,11 +51,7 @@ void FiberQueue::Run() {
     push_ec_.notifyAll();
 
     for (unsigned i = 0; i < count; ++i) {
-      try {
-        batch[i]();
-      } catch (std::exception& e) {
-        LOG(FATAL) << "Exception " << e.what();
-      }
+      batch[i]();
     }
```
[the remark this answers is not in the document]
romange: We declared Tasklet as no throw

util/fibers/fiberqueue_threadpool.cc
```diff
@@ -49,11 +51,7 @@ void FiberQueue::Run() {
     push_ec_.notifyAll();
 
     for (unsigned i = 0; i < count; ++i) {
-      try {
```
glevkovich, line 52: I think it is wrong to remove the try/catch.
It will still throw even with Tasklet - the isThrowing just mark how to act on  a specific case:
See function2.hpp:
```

/// \tparam IsThrowing Defines whether the function throws an exception on
///                    empty function call, `std::abort` is called otherwise.

```
The signature doesn't have noexcept at the end, so it can still throw. If you remove this try/catch the LOG(FATAL) won't log the error and any exception will be treated as an uncaught exception.

You can return this block back or add noexcept at the end of fu2Fun:
```
fu2::function_base<true, false, fu2::capacity_fixed<32, 8>,
                   false, false, void() noexcept>;
```

util/fibers/fiberqueue_threadpool.h
```diff
@@ -87,15 +108,26 @@ class FiberQueue {
 
   void Run();
 
+  unsigned blocked_submitters() const {
```
glevkovich, line 111: 1. Nit: This is dead code, why do you add it if no one uses it? I assume it is for future use? 
2. If you decide to keep this function, I think you should document it (add a comment above) because the name + non-static signature give the impression a per-queue counter, which it is not (it can be confusing) .
3. Also, if it stays consider making it static, since it reads thread-global state and has nothing to do with a specific instance - a non-static accessor makes caller think they can read it as per-queue.

util/fibers/fiberqueue_threadpool.h
```diff
@@ -52,8 +70,11 @@ class FiberQueue {
       if (TryAdd(std::forward<F>(f))) {
         break;
       }
+
       result = true;
+      blocked_submitters_++;
```
glevkovich, line 75: Shouldn't a "blocked submitter" be accounted also in line 176? 
I don't see `blocked_submitters_` being updated there.

util/fibers/fiberqueue_threadpool.h
```diff
@@ -19,6 +24,19 @@ class FiberQueueThreadPool;
...
+  using Fu2Fun =
+      fu2::function_base<true /* owns */, false /*non-copyable*/, fu2::capacity_fixed<32, 8>,
+                         false /* non-throwing*/, false /* strong exceptions guarantees*/, void()>;
+
+  struct Tasklet : public Fu2Fun {
+    using Fu2Fun::Fu2Fun;
+    using Fu2Fun::operator=;
+  };
+
+  static_assert(sizeof(Tasklet) == 48);
```
glevkovich, line 38: nit: consider adding a message here in case compilation fail.
e.g:
```
static_assert(sizeof(Tasklet) == 48,
              "Tasklet size changed: 32B inline buffer + fu2 dispatch overhead. ");
```

util/fibers/fiberqueue_threadpool.h
```diff
@@ -19,6 +24,19 @@ class FiberQueueThreadPool;
 class FiberQueue {
   friend class FiberQueueThreadPool;
 
+  // Fixed-capacity, non-copyable, non-throwing callable. Callbacks up to 32 bytes are stored
+  // inline (no heap allocation), unlike std::function. Keeps queue cells trivially relocatable.
```
glevkovich, line 28: nit:
I think that "Keeps queue cells trivially relocatable." is not correct.
From my understanding  "trivially relocatable" is true if we can move the object  to a new address using memcpy - but this is not what i see in function2.hpp. e.g:

```
  void move(vtable& to_table, data_accessor* from, std::size_t from_capacity,
            data_accessor* to,
            std::size_t to_capacity) noexcept(HasStrongExceptGuarantee) {
    cmd_(&to_table, opcode::op_move, from, from_capacity, to, to_capacity);
     ....
```
and after that  for inline storage that op_move path move-constructs the callable and destroys the source:

```
  // The object is allocated inplace
          else {
            construct(std::true_type{}, std::move(*box), to_table, to,
                      to_capacity);
            box->~T();
          }
```


I would just deleted "Keeps queue cells trivially relocatable."

util/fibers/fiberqueue_threadpool.h
```diff
@@ -87,15 +108,26 @@ class FiberQueue {
...
-  EventCount push_ec_, pull_ec_;
+  // pull_ec_ is written by every producer (notify() after each commit); push_ec_ is written by the
+  // consumer (notifyAll() once per drained batch) and by blocked producers. Keep these two on
+  // separate cache lines so producer-side and consumer-side wakeup traffic don't ping-pong the same
+  // line. is_closed_ needs no isolation: it is written once (at Shutdown) and only read on the
+  // consumer's park path, where it already touches push_ec_ anyway. queue_ isolates its own
+  // enqueue_pos_/dequeue_pos_ internally.
+  alignas(64) EventCount pull_ec_;
+  alignas(64) EventCount push_ec_;
   std::atomic_bool is_closed_{false};
```
BorysTheDev, line 128: What about is_closed_, it is in the same cache line as push_ec_ so if we read it we get push_ec_

BorysTheDev (approved):

util/fibers/fiberqueue_threadpool.h
romange: I checked it. Since `is_closed_` changes only once, it does not really matter. As long as variable does not change it does not affect cache-lines.

util/fibers/fiberqueue_threadpool.h
romange: yes, it should be static

PR #606, opened 2026-07-05
perf(fibers): avoid heap allocation for scoped ProactorPool blocking counters

romange:
## Summary
- Add `FiberBlockingCounter`, backed by an `EmbeddedBlockingCounter` embedded directly in `FiberInterface` instead of a heap-allocated `shared_ptr`. Lifetime is anchored to the owning fiber's existing intrusive_ptr refcount, so remote `Dec()`-calling closures keep the fiber (and its counter) alive through the trailing `notifyAll()` call, without any extra allocation.
- `ProactorPool::Await`/`AwaitBrief`/`AwaitFiberOnAll` now use `FiberBlockingCounter` since each is a tight, single-call, scoped rendezvous (construct → dispatch → `Wait()`, nothing else touches a blocking counter on that fiber in between).
- `BlockingCounter` (heap/`shared_ptr`-based) is unchanged and still used by long-lived owners like `AcceptServer::ref_bc_`, whose `Run()`/`Wait()` are split across separate calls - embedding those in the fiber's single slot would let unrelated code sharing that fiber corrupt an in-flight round (caught this exact crash in `accept_server_test` during development, see test plan).
- Split `EventCount`/`EmbeddedBlockingCounter` declarations out of `synchronization.h` into a new `util/fibers/detail/event_count.h` so `FiberInterface` can embed the counter without an include cycle.

## Test plan
- [x] Full project build (`ninja`) succeeds.
- [x] `fibers_test`, `accept_server_test` (all 14 cases), `uring_file_test`, `fiber_socket_test` pass.
- [x] Full `ctest` run: 23/24 pass; the one failure (`tls_socket_test`, kTLS negotiation) reproduces identically on unmodified `master` - confirmed pre-existing/environment-specific, unrelated to this change.
- [x] Regression-tested: an earlier version of this change made `BlockingCounter` itself embedded-backed, which crashed `accept_server_test` because `AcceptServer`'s `Run()`/`Wait()` split shares one fiber's slot with unrelated `AcceptServer` instances constructed on the same thread. Scoped the optimization to `FiberBlockingCounter`/`ProactorPool` instead.

## Summary by CodeRabbit

* **New Features**
  * Added improved fiber-based waiting and synchronization support, including a new lightweight blocking counter option.
  * Expanded completion tracking for pool waits, making fiber waits more efficient and consistent.

* **Bug Fixes**
  * Updated wait handling to better support fiber activity and reduce issues with blocking operations during asynchronous work.


util/fibers/event_count.h
```diff
@@ -0,0 +1,224 @@
+// Copyright 2023, Roman Gershman.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
```
romange, line 4: moved into this file

util/fibers/synchronization.h
```diff
@@ -413,75 +270,48 @@ class Done {
...
 };
 
-// A barrier similar to Go's WaitGroup for tracking remote tasks.
-// Internal smart pointer for easier lifetime management. Pass by value.
-class BlockingCounter {
+// Same semantics as BlockingCounter, but backed by the constructing fiber's embedded
+// EmbeddedBlockingCounter (see FiberInterface::blocking_counter()) instead of a heap
+// allocation. Pass by value: copies share the same underlying counter and keep the owning
+// fiber's FiberInterface (and thus the counter) alive via intrusive_ptr refcounting until
+// every copy - including ones captured by remote Dec()-calling closures - is destroyed.
```
dranikpg, line 295: did it show up in a profile? it seems like a thread-local allocation is really cheap and the shared_ptr has the same atomic operations

dranikpg (approved):

util/fibers/synchronization.h
romange: This is from the profile I made. So the improvement is not just around `BlockingCounter` allocation but also to tackle number 2: implicit heap allocation of lambda moved to fu2::function because it captures too many things. sizeof(BlockingCounter) == 16, sizeof(FiberBlockingCounter) is 8 so it will be easier to pack the lambda inline.

```
CPU25 shard-thread profile findings (recorded with -C 25):

This run looks like a real shard-core view, unlike the earlier CPU37 ingress-core profile.

Top shard-side hotspots:

Transaction::StoreKeysInArgs ~7.95%
squashed lambda invoke (fu2 ... internal_invoker<...lambda#2...>) ~5.43%
MultiCommandSquasher::SquashedHopCb ~5.01%
Transaction::MultiSwitchCmd ~3.86%
ProactorBase::Pause ~3.36%
mi_free_block_delayed_mt ~3.21%
EmbeddedBlockingCounter::Dec ~2.96%
Transaction::InitByKeys ~2.43%
Bucket::FindByFp ~2.04%
DashTable::FindFirst ~1.91%
mi_new ~1.90%
XXH64 ~1.53%
Transaction::GetDbSlice / Namespace::GetDbSlice ~1.53% / 1.74%
Transaction::ScheduleInShard ~1.52%
DbSlice::Acquire / LockTable::Release ~1.44% each
CmdGet ~1.39%
DbSlice::FindReadOnly ~1.24%
DbSlice::FindInternal ~0.53%
``

romange: it should also reduce `mi_free_block_delayed_mt` as lambda is allocated in one thread and freed in another.

dranikpg: ah, reducing size is also a valid optimization point, right

PR #609, opened 2026-07-08
Fibers: consume FiberBlockingCounter callbacks

romange:
## Summary
- add consumable FiberBlockingCounter::Release() for callback-held copies
- update ProactorPool await helpers to use Release() instead of pointer-style Dec()
- clear cached proactor Tasklets after execution and add a refcount regression test

## Test
- pre-commit run --files util/fibers/epoll_proactor.cc util/fibers/fibers_test.cc util/fibers/synchronization.cc util/fibers/synchronization.h util/fibers/uring_proactor.cc util/proactor_pool.h
- ./fibers_test --gtest_filter=FiberTest.ProactorPoolRunDoesNotRetainMainFiberRef --logtostderr

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of fiber/task resources after work completes, reducing the chance of lingering references.
  * Updated fiber synchronization so waiting and completion handling are more explicit and reliable.
  * Added coverage for proactor pool behavior to verify fibers are not retained after running or stopping.
  * Kept existing task dispatch behavior the same while tightening internal resource release timing.

### DIFF
--- a/util/fibers/synchronization.h
+++ b/util/fibers/synchronization.h
@@ -5,12 +5,12 @@
 #pragma once
 
 #include <atomic>
+#include <boost/smart_ptr/intrusive_ptr.hpp>
 #include <cassert>
 #include <condition_variable>  // for cv_status
 #include <memory>
 #include <optional>
-
-#include <boost/smart_ptr/intrusive_ptr.hpp>
+#include <utility>
 
 #include "base/spinlock.h"
 #include "util/fibers/detail/fiber_interface.h"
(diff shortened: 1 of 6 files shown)


util/fibers/synchronization.h
```diff
@@ -304,10 +304,13 @@ class FiberBlockingCounter {
  public:
   explicit FiberBlockingCounter(unsigned start_count);
 
-  EmbeddedBlockingCounter* operator->() {
-    return &owner_->blocking_counter();
+  bool Wait() {
+    return owner_->blocking_counter().Wait();
   }
 
+  void Release() &&;
```
kostasrim, line 311: That's actually a nice idiom to enforce the ref count drop on `dec`

kostasrim (approved):
lgtm

PR #611, opened 2026-07-12
fix(epoll): bound completion batches to prevent scheduler starvation

romange:
## Summary
Fixes the regression that was introduced by #603.

Process at most `kMaxCompletionBatches`  full epoll batches before yielding to the scheduler, preventing level-triggered FDs from starving the fibers that consume them.

## Changes
- Bound nonblocking epoll completion draining to kMaxCompletionBatches full batches.
- Document the level-triggered readiness livelock scenario.
- Add a 128-eventfd regression in `ProactorTest` that requires multiple batches and scheduler progress.
- Fix MacOs epoll behavior that was broken unrelated to #603.

### DIFF
--- a/util/fibers/epoll_proactor.cc
+++ b/util/fibers/epoll_proactor.cc
@@ -62,7 +62,7 @@ int EpollWait(int epoll_fd, EventsBatch* batch, int timeout) {
   return epoll_wait(epoll_fd, batch->cqe, kEvBatchSize, timeout);
 }
 
-void EpollDel(int epoll_fd, int fd) {
+void EpollDel(int epoll_fd, int fd, uint32_t /*event_mask*/) {
   CHECK_EQ(0, epoll_ctl(epoll_fd, EPOLL_CTL_DEL, fd, NULL));
 }
 
@@ -89,20 +89,24 @@ int EpollCreate() {
 }
 
 int EpollWait(int epoll_fd, EventsBatch* batch, int tm_ms) {
-  struct timespec ts {
-    .tv_sec = tm_ms / 1000, .tv_nsec = (tm_ms % 1000) * 1000000
-  };
+  struct timespec ts{.tv_sec = tm_ms / 1000, .tv_nsec = (tm_ms % 1000) * 1000000};
 
   int epoll_res = kevent(epoll_fd, NULL, 0, batch->cqe, kEvBatchSize, tm_ms < 0 ? NULL : &ts);
   return epoll_res;
 }
 
-void EpollDel(int epoll_fd, int fd) {
-  struct kevent kev[2];
+void EpollDel(int epoll_fd, int fd, uint32_t event_mask) {
+  if (event_mask & EpollProactor::EPOLL_IN) {
+    struct kevent kev;
+    EV_SET(&kev, fd, EVFILT_READ, EV_DELETE, 0, 0, NULL);
+    CHECK_EQ(0, kevent(epoll_fd, &kev, 1, NULL, 0, NULL));
+  }
 
-  EV_SET(&kev[0], fd, EVFILT_READ, EV_DELETE, 0, 0, NULL);
-  EV_SET(&kev[1], fd, EVFILT_WRITE, EV_DELETE, 0, 0, NULL);
-  CHECK_EQ(0, kevent(epoll_fd, kev, 2, NULL, 0, NULL));
+  if (event_mask & EpollProactor::EPOLL_OUT) {
+    struct kevent kev;
+    EV_SET(&kev, fd, EVFILT_WRITE, EV_DELETE, 0, 0, NULL);
+    CHECK_EQ(0, kevent(epoll_fd, &kev, 1, NULL, 0, NULL));
+  }
 }
 
 uint32_t KevMask(const struct kevent& kev) {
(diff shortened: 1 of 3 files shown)


romange:
qodo review again

util/fibers/epoll_proactor.cc
```diff
@@ -281,18 +281,31 @@ void EpollProactor::MainLoop(detail::Scheduler* scheduler) {
...
+      constexpr unsigned kMaxCompletionBatches = 3;
+
+      // Level-triggered FDs stay ready until a scheduled fiber consumes them, therefore
+      // we can not iterate endlessly without yielding to the scheduler.
+      // Break in two cases: a) we iterated kMaxCompletionBatches times,
+      // b) we got less than kEvBatchSize completions as an indication that we drained the queue.
+
+      // (b) by itself is not enough because there could be many level-triggered events
+      // that will fill ev_batch in full and we will never break.
+      for (unsigned batch = 0; batch < kMaxCompletionBatches; ++batch) {
```
glevkovich, line 297: I think there is a  bug here.
The socket is edge triggered, I looked at `void EpollSocket::OnSetProactor() {`
with:
`constexpr int kEventMask = EPOLLIN | EPOLLOUT | EPOLLET | EPOLLRDHUP;`

Lets assume you have 400 ready fds.
batch 0, 1,2 iterations execute -> at the end of batch 2 iteration:
`epoll_res = EpollWait(epoll_fd_, &ev_batch, 0);`
returns 16 more  ready fds, and loop condition breaks at the top.
Since you re-use a single ev_batch, and kernel remove them from the ready-list (since edge-triggered ) these ready are back to the list only if more bytes arrived.  in some cases nothing will happen else than performance degradation, and in other cases we will have a client hang.
This is a lost-wake up.

util/fibers/fibers_test.cc
```diff
@@ -1130,6 +1129,72 @@ TEST_P(ProactorTest, FiberFile) {
...
+      fiber_finished.Notify();
+    }).Detach();
+
+    uint64_t val = 1;
+    for (int fd : event_fds) {
+      EXPECT_EQ(8, write(fd, &val, sizeof(val)));
+    }
+  });
+
+  bool success = fiber_finished.WaitFor(chrono::seconds(1));
```
glevkovich: I think this cleanup (to line 1192) races with ReadyFdDrainFb: on timeout we close() the fds while the fiber may still be in read()/Disarm() on the proactor thread. Better make the fiber joinable (drop Detach()) and Join() before closing - and check cancelled inside the drain loop so the join can't hang.

util/fibers/fibers_test.cc
```diff
@@ -1130,6 +1129,72 @@ TEST_P(ProactorTest, FiberFile) {
   });
 }
 
+TEST_P(ProactorTest, ReadyFdsDoNotStarveScheduler) {
```
glevkovich, line 1137: This test  couldn't catch https://github.com/romange/helio/pull/611/changes#r3568131947

consider testing with #fds > 128*3 .

util/fibers/fibers_test.cc
```diff
@@ -1130,6 +1129,72 @@ TEST_P(ProactorTest, FiberFile) {
...
+  if (GetProactorType() == "uring") {
+    GTEST_SKIP() << "Skipped FiberSocketTest.ReadyFdsDoNotStarveScheduler test on uring";
```
glevkovich: message is incorrect
test is 
`ProactorTest.ReadyFdsDoNotStarveScheduler`
not
` FiberSocketTest.ReadyFdsDoNotStarveScheduler`

util/fibers/epoll_proactor.cc
```diff
@@ -89,20 +89,24 @@ int EpollCreate() {
...
-  };
+  struct timespec ts{.tv_sec = tm_ms / 1000, .tv_nsec = (tm_ms % 1000) * 1000000};
 
   int epoll_res = kevent(epoll_fd, NULL, 0, batch->cqe, kEvBatchSize, tm_ms < 0 ? NULL : &ts);
   return epoll_res;
 }
 
-void EpollDel(int epoll_fd, int fd) {
-  struct kevent kev[2];
+void EpollDel(int epoll_fd, int fd, uint32_t event_mask) {
```
romange, line 98: MacOs variant

glevkovich (approved):

glevkovich (dismissed):
approved with 4 comments, I see that some of them were addressed before submitted.

This small one still pending:
https://github.com/romange/helio/pull/611#discussion_r3568203320

util/fibers/epoll_proactor.cc
romange: if you are talking about breaking without dispatching - it has been fixed and I added a test coverage for it.

util/fibers/fibers_test.cc
romange: I think the motivation is to avoid deadlocks that come from a broken scheduler.  We need the main thread to guard for this and we can not rely on a fiber

util/fibers/fibers_test.cc
romange: removed

glevkovich (dismissed):
LGTM

glevkovich (approved):

PR #616, opened 2026-07-17
AWS: refresh expired IRSA S3 credentials

### LINKED ISSUE #7666: S3 snapshots are not working in 1.39
**Describe the bug**

I have clusters that were running version 1.38. After node maintenance, the operator automatically updated the version to 1.39, and the pods could no longer start.

They fail with the following error:

```
E0619 15:42:20.365695      10 utils.cc:255] Unexpected response HTTP/1.1 400 Bad Request
x-amz-bucket-region: eu-south-1
x-amz-request-id: E847D…
x-amz-id-2: GF0QIfLbrpC7FhJ8…
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Fri, 19 Jun 2026 15:42:19 GMT
Connection: close
Server: AmazonS3

47a
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidToken</Code><Message>The provided token is malformed or otherwise invalid.</Message><Token-0>FwoGZXI…</HostId></Error>
0


<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidToken</Code><Message>The provided token is malformed or otherwise invalid.</Message><Token-0>FwoGZXIvYX…</HostId></Error>

E0619 15:42:20.383545       7 server_family.cc:1326] Failed to load snapshot with error: end of stream: Failed list objects in S3 bucket
```

**To Reproduce**
1. Create a ServiceAccount with an IRSA role that has access to S3.
1. Deploy a cluster via the operator.
1. Upgrade or allow the operator to upgrade the cluster from 1.38 to 1.39.
1. …
1. Observe that the pods fail to start.

**Expected behavior**

Pods should start successfully after the upgrade and should be able to access the S3 bucket using the IRSA-provided credentials.

Rollback to 1.38.1 works as expected.

romange:
## Summary

- Refresh expired temporary AWS credentials before signing S3 requests.
- Serialize credential refreshes with a bounded, fiber-yielding wait and re-sign requests when credentials change.
- Treat S3 HTTP 400 `ExpiredToken` and `InvalidToken` as AWS-provider refresh-and-retry conditions.
- Add fake-S3 coverage for the expired-token retry path.

Fixes https://github.com/dragonflydb/dragonfly/issues/7666

## Tests

- `build-dbg/aws_creds_provider_test --logtostderr`
- `pytest -q tests/test_s3.py -k "upload_part or expired_token"`

### DIFF
--- a/util/cloud/aws/aws_creds_provider.cc
+++ b/util/cloud/aws/aws_creds_provider.cc
@@ -16,6 +16,7 @@
 
 #include <boost/beast/http/empty_body.hpp>
 #include <boost/beast/http/string_body.hpp>
+#include <chrono>
 #include <pugixml.hpp>
 
 #include "base/logging.h"
@@ -37,6 +38,8 @@ namespace {
 // SHA256("") — used as payload hash for all GET requests.
 constexpr string_view kEmptyBodyHash =
     "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855";
+constexpr auto kRefreshWaitTimeout = chrono::seconds(30);
+constexpr auto kRefreshWaitInterval = chrono::milliseconds(100);
 
 string HexEncode(const uint8_t* data, size_t len) {
   static const char kHex[] = "0123456789abcdef";
@@ -259,6 +262,7 @@ AwsCredsProvider::~AwsCredsProvider() {
 }
 
 SSL_CTX* AwsCredsProvider::GetSslCtx() {
+  folly::RWSpinLock::WriteHolder lock(lock_);
   if (!ssl_ctx_)
     ssl_ctx_ = http::TlsClient::CreateSslContext();
   return ssl_ctx_;
@@ -338,8 +342,8 @@ error_code AwsCredsProvider::TryEnvironment() {
   {
     folly::RWSpinLock::WriteHolder lock(lock_);
     creds_ = std::move(creds);
+    source_ = CredSource::kEnv;
   }
-  source_ = CredSource::kEnv;
   LOG_FIRST_N(INFO, 1) << "aws: loaded credentials; provider=environment";
   return {};
 }
@@ -382,8 +386,8 @@ error_code AwsCredsProvider::TryProfileFile() {
   {
     folly::RWSpinLock::WriteHolder lock(lock_);
     creds_ = std::move(creds);
+    source_ = CredSource::kProfile;
   }
-  source_ = CredSource::kProfile;
   LOG_FIRST_N(INFO, 1) << "aws: loaded credentials; provider=profile file";
   return {};
 }
@@ -420,10 +424,8 @@ error_code AwsCredsProvider::TryWebIdentity() {
   {
     folly::RWSpinLock::WriteHolder lock(lock_);
     creds_ = std::move(creds);
+    source_ = CredSource::kWebIdentity;
   }
-  role_arn_ = role_arn;
-  web_identity_token_file_ = token_file;
-  source_ = CredSource::kWebIdentity;
   LOG_FIRST_N(INFO, 1) << "aws: loaded credentials; provider=web-identity";
   return {};
 }
@@ -452,8 +454,6 @@ error_code AwsCredsProvider::TryContainerCreds() {
     }
   }
 
-  container_uri_ = uri;
-
   // Determine if http or https.
   vector<pair<string, string>> headers;
   if (!auth_token.empty()) {
@@ -472,8 +472,8 @@ error_code AwsCredsProvider::TryContainerCreds() {
   {
     folly::RWSpinLock::WriteHolder lock(lock_);
     creds_ = std::move(creds);
+    source_ = CredSource::kContainer;
   }
-  source_ = CredSource::kContainer;
   LOG_FIRST_N(INFO, 1) << "aws: loaded credentials; provider=container-credentials";
   return {};
 }
@@ -495,17 +495,16 @@ error_code AwsCredsProvider::TryIMDS() {
                h2::verb::get, token_hdr, "", connect_ms_, nullptr);
   if (!role_name_res)
     return role_name_res.error();
-  string role_name_raw = std::move(*role_name_res);
+  string imds_role(absl::StripAsciiWhitespace(*role_name_res));
 
-  imds_role_ = string(absl::StripAsciiWhitespace(role_name_raw));
-  if (imds_role_.empty()) {
+  if (imds_role.empty()) {
     LOG(WARNING) << "aws: IMDS returned empty role name";
     return make_error_code(errc::not_supported);
   }
 
   // Get the credentials for the role.
   auto json = FetchUrl("169.254.169.254", "80",
-                       absl::StrCat("/latest/meta-data/iam/security-credentials/", imds_role_),
+                       absl::StrCat("/latest/meta-data/iam/security-credentials/", imds_role),
                        h2::verb::get, token_hdr, "", connect_ms_, nullptr);
   if (!json)
     return json.error();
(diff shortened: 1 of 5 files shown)


util/cloud/aws/aws_creds_provider.cc
vyavdoshenko: source_, role_arn_, web_identity_token_file_, container_uri_ and imds_role_ are written outside lock_ (the lock only guards creds_). The provider is shared across shards, and this PR makes concurrent RefreshToken() the common case (proactive pre-loop refresh + 400 ExpiredToken recognition on every credential rotation). Example: imds_role_ is written at line 500 and read back at line 508, so two concurrent TryIMDS() calls give a torn read / crash; for the string fields a write-write copy-assign can double-free (UB, TSan-detectable). Please put these assignments under the write lock (and read source_ under the read lock in RefreshToken), or add single-flight (see the IsExpired() comment). Same issue at lines 455, 476, 500/508 and 520.

util/cloud/utils.cc
```diff
@@ -173,6 +178,12 @@ ParsedHttpUrl ParseHttpUrl(string_view uri) {
 error_code RobustSender::Send(unsigned num_iterations, detail::HttpRequestBase* req,
                               SenderResult* result) {
   error_code ec;
+  if (provider_->IsExpired()) {
+    VLOG(1) << "Refreshing expired token before sending request";
+    RETURN_ERROR(provider_->RefreshToken());
```
vyavdoshenko: IsExpired() fires 30s BEFORE real expiry, while the cached credentials are still valid. This RETURN_ERROR aborts the whole request before it is ever sent if RefreshToken() fails transiently (STS 5xx/throttle) - yet pre-PR this same request went out with the still-valid creds and succeeded. That is a regression on the exact IRSA path this PR fixes. Suggest making proactive refresh best-effort: on failure, log and proceed with the current creds (a genuine expiry is still caught by the reactive 400 ExpiredToken path in the loop). Keep the hard fail only when the creds are actually past expiry.

util/cloud/utils.cc
```diff
@@ -173,6 +178,12 @@ ParsedHttpUrl ParseHttpUrl(string_view uri) {
 error_code RobustSender::Send(unsigned num_iterations, detail::HttpRequestBase* req,
                               SenderResult* result) {
   error_code ec;
+  if (provider_->IsExpired()) {
```
vyavdoshenko: IsExpired() is a pure clock check against the shared creds_, so every shard flips to true on the same tick and each independently fires its own AssumeRoleWithWebIdentity - a synchronized burst of identical STS refreshes with no coalescing, which raises the chance of STS throttling (and feeds the RETURN_ERROR abort above). Please add single-flight: an fb2::Mutex plus a re-check of IsExpired() after acquiring it, so the first caller refreshes and the rest reuse the result. This also closes the data race on the unlocked provider fields.

romange:
wow, is it opus-4.8 that found all these @vyavdoshenko ?

vyavdoshenko:
> wow, is it opus-4.8 that found all these @vyavdoshenko ?

yes, and double-checked with gpt 5.5

util/cloud/utils.cc
romange: I reduced 30s to 1s to simplify the semantics. We could have preemprive refresh but these should not block the main request anyways.

util/cloud/aws/aws_creds_provider.cc
romange: Addressed in 20cfe5e. Refresh-source reads and credential/source writes now use the same RW lock; unused shared refresh metadata was removed, IMDS role names are request-local, and lazy TLS-context initialization is locked.

kostasrim:
> > wow, is it opus-4.8 that found all these @vyavdoshenko ?
> 
> yes, and double-checked with gpt 5.5

if we are doing opus/gpt reviews now why don't we add a hook so we don't have to copy paste prompts ?

kostasrim:
> > > wow, is it opus-4.8 that found all these @vyavdoshenko ?
> > 
> > 
> > yes, and double-checked with gpt 5.5
> 
> if we are doing opus/gpt reviews now why don't we add a hook so we don't have to copy paste prompts ?

In fact we can:

Trigger on PR events (opened, synchronize, reopened) in GitHub Actions.
A GitHub Action gathers the PR diff.
The action calls the Anthropic API with your review prompt.
The action posts the review back to the PR (either as a review or comments).

Then something like 

```
name: AI PR Review

on:
  pull_request:
    types: [opened, synchronize, reopened]

jobs:
  review:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0

      - run: |
          git fetch origin ${{ github.event.pull_request.base.ref }}
          git diff origin/${{ github.event.pull_request.base.ref }}...HEAD > diff.txt

      - run: python review.py
        env:
          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}****
```

If you think that would be interesting I could follow up.

romange:
I think our current PR agents should provide these feedbacks. If they do not give good results, I wanna know

romange:
qodo review

kostasrim:
> I think our current PR agents should provide these feedbacks. If they do not give good results, I wanna know

But the details from one review to the other are night and day 🤷‍♂️  One was in-depth and on point while the PR agents did not even scratched it. I am not sure -- just thinking out loud here on how to improve our processes and at least have some sense of safety that does not rely on people running manually models (not because I have a problem with it but because I want to automate this and not rely on human attention/error). Otherwise we just compete on who runs what ...

util/cloud/utils.cc
romange: actually changed this back to 30s

romange:
augment review

util/cloud/aws/aws_creds_provider.cc
```diff
@@ -516,14 +515,40 @@ error_code AwsCredsProvider::TryIMDS() {
...
+  // Someone else is already refreshing; wait for it to finish.
+  auto deadline = chrono::steady_clock::now() + kRefreshWaitTimeout;
+  while (is_refreshing_.load(memory_order_acquire)) {
+    if (chrono::steady_clock::now() >= deadline) {
+      return make_error_code(errc::timed_out);
+    }
+    ThisFiber::SleepFor(kRefreshWaitInterval);
+  }
+
+  return {};
```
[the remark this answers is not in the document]
romange: it's fine as the result will be passed to the first call that was finished, and retrying again does not break correctness

util/cloud/aws/aws_creds_provider.cc
```diff
@@ -641,5 +662,46 @@ void AwsCredsProvider::Sign(detail::HttpRequestBase* req) const {
...
+    {
+      folly::RWSpinLock::ReadHolder creds_lock(lock_);
+      if (!creds_.IsExpiring())
+        return false;
+    }
+
+    bool expected = false;
+    if (is_refreshing_.compare_exchange_strong(expected, true, memory_order_acq_rel)) {
+      error_code ec = RefreshTokenImpl();
+      is_refreshing_.store(false, memory_order_release);
```
vyavdoshenko, line 677: is_refreshing_ is reset with a bare store(false). If RefreshTokenImpl() throws (e.g. bad_alloc, or fiber cancellation during the network I/O), this store is skipped and the flag stays true forever: every later RefreshToken()/RefreshIfExpiring() becomes a waiter that times out after 30s, credentials never refresh again, and every Send eats a 30s stall. Please clear the flag via RAII (absl::Cleanup / a scope guard) so it is reset on every exit path.

util/cloud/aws/aws_creds_provider.cc
```diff
@@ -516,14 +515,40 @@ error_code AwsCredsProvider::TryIMDS() {
...
+  LOG_FIRST_N(INFO, 1) << "aws: loaded credentials; provider=ec2-metadata; role=" << imds_role;
   return {};
 }
 
 error_code AwsCredsProvider::RefreshToken() {
-  switch (source_) {
+  bool expected = false;
+  if (is_refreshing_.compare_exchange_strong(expected, true, memory_order_acq_rel)) {
+    error_code ec = RefreshTokenImpl();
+    is_refreshing_.store(false, memory_order_release);
```
vyavdoshenko, line 528: if RefreshTokenImpl() throws, is_refreshing_ is never cleared and refresh wedges permanently. Reset it via RAII instead of a bare store(false).

vyavdoshenko (approved):

util/cloud/aws/aws_creds_provider.cc
romange: 1. this is true with 95% of our code. the assumption is that our code does not throw :)

2. the comment is AI slope, helio fiber can not be cancelled.

PR #623, opened 2026-07-29
refactor(tls): encapsulate socket state flags

glevkovich:
- Replace C-style state-byte masking with SocketFlags, providing clear predicates, setters, clearers, and wait helpers for socket state.
- Route blocking, non-blocking, and async TLS operations through the shared API, preserving the existing state-bit values and behavior.

### DIFF
--- a/util/tls/tls_socket.h
+++ b/util/tls/tls_socket.h
@@ -245,9 +245,6 @@ class TlsSocket final : public FiberSocketBase {
     void set_shutdown_in_progress() {
       state_ |= SHUTDOWN_IN_PROGRESS;
     }
-    void set_shutdown_done() {
-      state_ |= SHUTDOWN_DONE;
-    }
     void set_user_recv_in_progress() {
       state_ |= USER_RECV_IN_PROGRESS;
     }
@@ -256,9 +253,6 @@ class TlsSocket final : public FiberSocketBase {
     }
 
     // Clearers
-    void clear_shutdown_in_progress() {
-      state_ &= ~SHUTDOWN_IN_PROGRESS;
-    }
     void clear_user_recv_in_progress() {
       state_ &= ~USER_RECV_IN_PROGRESS;
     }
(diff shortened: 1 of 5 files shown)


util/tls/tls_socket.h
```diff
@@ -204,7 +197,99 @@ class TlsSocket final : public FiberSocketBase {
...
   };
-  uint8_t state_{0};
+
+  // Shared state and synchronization for the blocking, non-blocking, and async paths.
+  class SocketFlags {
+   public:
+    // Test-only whole-state replacement.
+    void overwrite(uint8_t state) {
+      state_ = state;
+    }
```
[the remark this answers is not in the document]
glevkovich: fixed

util/tls/tls_socket.h
```diff
@@ -204,7 +197,99 @@ class TlsSocket final : public FiberSocketBase {
     // A recv-path background write draining the engine's output is in flight.
     RECV_DRAIN_ENGINE_IN_FLIGHT = 1 << 5,
   };
-  uint8_t state_{0};
+
+  // Shared state and synchronization for the blocking, non-blocking, and async paths.
+  class SocketFlags {
+   public:
+    // Test-only whole-state replacement.
+    void overwrite(uint8_t state) {
```
kostasrim, line 205: Don't we break the style guide here ? e.g. bits() vs Bits() ?

kostasrim: same for all the functions in this file

kostasrim (commented):
lgtm, just take care of the naming of the members 

IMO my preference is to keep functions small but avoid deep nested calls (just single line functions that call other functions) but that's a preference I guess

romange (approved):

util/tls/tls_socket.h
glevkovich: I did not break, and of  that was done while being aware of the lower-case - as these are mainly getters/setters/clearers:

https://google.github.io/styleguide/cppguide.html#Function_Names


`Accessors and mutators (get and set functions) may be named like variables, in snake_case. These often correspond to actual member variables, but this is not required. For example, int count() and void set_count(int count).`

PR #624, opened 2026-07-29
chore: cancel in-flight requests for iouring and epoll socket

kostasrim:
* allow cancelling in-flight epoll and iouring read/writes
* add tests to cover all paths

### DIFF
--- a/util/fiber_socket_base.h
+++ b/util/fiber_socket_base.h
@@ -199,6 +199,14 @@ class LinuxSocketBase : public FiberSocketBase {
 
   error_code Shutdown(int how) override;
 
+  // Cancels all pending async operations (reads and writes) submitted on this socket's fd,
+  // forcing them to fail immediately instead of waiting to resolve naturally. Unlike
+  // Shutdown()/Close(), this can interrupt an operation that is already in flight - e.g. a
+  // write stalled because the peer isn't draining its receive buffer, which Shutdown()/Close()
+  // cannot force to complete, since io_uring holds its own reference to a submitted operation
+  // until it resolves on its own.
+  virtual error_code CancelInFlightOps() = 0;
+
   // UINT32_MAX to disable timeout.
   void set_timeout(uint32_t msec) final override {
     timeout_ = msec;
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -193,6 +193,47 @@ auto EpollSocket::Close() -> error_code {
   return ec;
 }
 
+std::error_code EpollSocket::CancelInFlightOps() {
+  // Cleanup pending requests. Unlike UringSocket, there's no kernel-held reference to interrupt
+  // here - write_req_/async_write_req_ (and read_req_/async_read_req_/on_recv_) are a union:
+  // at most one of {synchronous WriteSome() suspended via PendingReq, a queued AsyncWriteSome(),
+  // <read side: also a registered OnRecv hook>} can be outstanding at a time. So cancelling is
+  // just a matter of resolving whichever one is currently set ourselves.
+  //
+  // Unlike Close()'s own cleanup above, this also wakes a fiber suspended in the synchronous
+  // WriteSome()/RecvMsg() path - that's the whole point of this method (see CancelInFlightOps()
+  // on UringSocket, which cancels in-flight kernel ops regardless of whether they were
+  // submitted synchronously or asynchronously). It is intentionally not called from Close().
+  if (async_write_pending_) {
+    DCHECK(async_write_req_);
+    async_write_req_->cb(MakeUnexpected(errc::operation_canceled));
+    delete async_write_req_;
+    async_write_req_ = nullptr;
+    async_write_pending_ = 0;
+  } else if (write_req_ && write_req_->IsSuspended()) {
+    // A fiber is blocked inside the synchronous WriteSome() - wake it with an error instead of
+    // leaving it suspended forever.
+    write_req_->Activate(make_error_code(errc::operation_canceled));
+  }
+
+  if (async_read_pending_) {
+    DCHECK(async_read_req_);
+    async_read_req_->cb(MakeUnexpected(errc::operation_canceled));
+    delete async_read_req_;
+    async_read_req_ = nullptr;
+    async_read_pending_ = 0;
+  } else if (recv_hook_registered_) {
+    delete on_recv_;
+    on_recv_ = nullptr;
+    recv_hook_registered_ = 0;
+  } else if (read_req_ && read_req_->IsSuspended()) {
+    // A fiber is blocked inside the synchronous RecvMsg() - wake it with an error instead of
+    // leaving it suspended forever.
+    read_req_->Activate(make_error_code(errc::operation_canceled));
+  }
+  return {};
+}
+
 void EpollSocket::OnSetProactor() {
   if (fd_ >= 0) {
     CHECK_LT(arm_index_, 0);
--- a/util/fibers/epoll_socket.cc
+++ b/util/fibers/epoll_socket.cc
@@ -204,24 +204,34 @@ std::error_code EpollSocket::CancelInFlightOps() {
   // WriteSome()/RecvMsg() path - that's the whole point of this method (see CancelInFlightOps()
   // on UringSocket, which cancels in-flight kernel ops regardless of whether they were
   // submitted synchronously or asynchronously). It is intentionally not called from Close().
+  LOG(WARNING) << "diag: CancelInFlightOps entry, async_write_pending_=" << int(async_write_pending_)
+               << " write_req_=" << write_req_
+               << " suspended=" << (write_req_ ? write_req_->IsSuspended() : -1);
   if (async_write_pending_) {
     DCHECK(async_write_req_);
-    async_write_req_->cb(MakeUnexpected(errc::operation_canceled));
-    delete async_write_req_;
+    AsyncReq* req = async_write_req_;
+    auto cb = std::move(req->cb);
+    delete req;
     async_write_req_ = nullptr;
     async_write_pending_ = 0;
+    cb(MakeUnexpected(errc::operation_canceled));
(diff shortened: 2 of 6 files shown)


util/fibers/epoll_socket.cc
```diff
@@ -193,6 +193,47 @@ auto EpollSocket::Close() -> error_code {
...
+  // <read side: also a registered OnRecv hook>} can be outstanding at a time. So cancelling is
+  // just a matter of resolving whichever one is currently set ourselves.
+  //
+  // Unlike Close()'s own cleanup above, this also wakes a fiber suspended in the synchronous
+  // WriteSome()/RecvMsg() path - that's the whole point of this method (see CancelInFlightOps()
+  // on UringSocket, which cancels in-flight kernel ops regardless of whether they were
+  // submitted synchronously or asynchronously). It is intentionally not called from Close().
+  if (async_write_pending_) {
+    DCHECK(async_write_req_);
+    async_write_req_->cb(MakeUnexpected(errc::operation_canceled));
```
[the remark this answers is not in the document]
kostasrim: we exit with cancelled it should not happen

util/fibers/fiber_socket_test.cc
```diff
@@ -940,5 +940,109 @@ TEST_P(FiberSocketTest, LeakAsyncWrite) {
...
+      write_ec = ec;
+      write_completed = true;
+      write_done.Notify();
+    });
+  });
+
+  // The write must still be stuck - nobody is reading conn_socket_.
+  EXPECT_FALSE(write_done.WaitFor(100ms))
+      << "write should not have completed - nobody is draining the peer";
+  EXPECT_FALSE(write_completed);
```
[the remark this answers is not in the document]
kostasrim: 😮‍💨 In theory yes in practice NO. Why ? Because we are not reading on the other side. Therefore the write will never make progress so it's safe because there is no parallelism involved at all. They can never race because the in-flight write can not make progress...

util/fibers/fiber_socket_test.cc
```diff
@@ -940,5 +940,109 @@ TEST_P(FiberSocketTest, LeakAsyncWrite) {
...
+  // Write() suspends the calling fiber until everything is sent, so run it on its own fiber and
+  // observe it from here.
+  Fiber write_fb = proactor_->LaunchFiber("BlockedWrite", [&] {
+    write_ec = sock->Write(io::Bytes(buf.get(), kBufSize));
+    write_completed = true;
+  });
+
+  // The write must still be stuck - nobody is reading conn_socket_.
+  ThisFiber::SleepFor(100ms);
+  EXPECT_FALSE(write_completed) << "write should not have completed - nobody is draining the peer";
```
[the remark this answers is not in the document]
kostasrim: Again same principles, the write is blocked and it will never make progress as the other side is not reading...

util/fibers/uring_socket.cc
```diff
@@ -103,6 +103,35 @@ auto UringSocket::Close() -> error_code {
...
+                   IORING_ASYNC_CANCEL_ALL;
+
+  // CancelRequests() wraps io_uring_register_sync_cancel(), which is a real synchronous
+  // syscall (IORING_REGISTER_SYNC_CANCEL) - not a submit-then-reap-completion SQE like every
+  // other operation here. It blocks the calling thread in the kernel while it locates and
+  // cancels matching requests (not a userspace busy-loop), but that also means it does NOT
+  // yield to other fibers on this proactor the way normal io_uring ops do - the whole thread
+  // is blocked for its duration. Should be fine in practice since cancellation itself is a fast
+  // kernel-side lookup+abort, not I/O and it's a rather "exceptional" path.
+  int res = GetProactor()->CancelRequests(ShiftedFd(), flags);
```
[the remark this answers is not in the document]
kostasrim: good one. reconcile it. will fix soon

kostasrim:
@coderabbitai review

util/fibers/epoll_socket.cc
```diff
@@ -193,6 +193,57 @@ auto EpollSocket::Close() -> error_code {
...
+  // here - write_req_/async_write_req_ (and read_req_/async_read_req_/on_recv_) are a union:
+  // at most one of {synchronous WriteSome() suspended via PendingReq, a queued AsyncWriteSome(),
+  // <read side: also a registered OnRecv hook>} can be outstanding at a time. So cancelling is
+  // just a matter of resolving whichever one is currently set ourselves.
+  //
+  // Unlike Close()'s own cleanup above, this also wakes a fiber suspended in the synchronous
+  // WriteSome()/RecvMsg() path - that's the whole point of this method (see CancelInFlightOps()
+  // on UringSocket, which cancels in-flight kernel ops regardless of whether they were
+  // submitted synchronously or asynchronously). It is intentionally not called from Close().
+  LOG(WARNING) << "diag: CancelInFlightOps entry, async_write_pending_=" << int(async_write_pending_)
```
glevkovich, line 207: Is this a leftover? not sure how many times this function is going to be called but if its called a lot it will be very noise in log.
Make it DVLOG(2) or completely remove.


Same for lines 221,223

util/fibers/fiber_socket_test.cc
```diff
@@ -940,5 +940,169 @@ TEST_P(FiberSocketTest, LeakAsyncWrite) {
...
+
+  constexpr size_t kBufSize = 16 * 1024 * 1024;  // Exceeds any realistic kernel socket buffer.
+  std::unique_ptr<uint8_t[]> buf(new uint8_t[kBufSize]);
+
+  error_code write_ec;
+  bool write_completed = false;
+
+  // Write() suspends the calling fiber until everything is sent, so run it on its own fiber and
+  // observe it from here.
+  LOG(WARNING) << "diag: launching write_fb";
```
glevkovich, line 1026: This test has 9 LOG(WARNING). Now that you completed write it, I suggest to remove them all.

util/fibers/epoll_socket.cc
glevkovich, line 449: you should add here same like in WriteSome ECANCELED, else it will print an error.

util/fiber_socket_base.h
```diff
@@ -199,6 +199,14 @@ class LinuxSocketBase : public FiberSocketBase {
 
   error_code Shutdown(int how) override;
 
+  // Cancels all pending async operations (reads and writes) submitted on this socket's fd,
```
glevkovich, line 202: You mention here that this function cancels pending reads and writes, but it also silently removes callbacks registered with `RegisterOnRecv()`. On `io_uring`, it also removes the callback from `RegisterOnErrorCb()`, while `epoll` keeps it alive. My main concern is the backend undocumented part: after calling `CancelInFlightOps()`, an error callback might still fire under `epoll`, but will never fire under `io_uring`.

Please add documentation for the shared behavior here - add the removal of RegisterOnRecv -  and add backend-specific details next to each `CancelInFlightOps()` override (in the uring and epoll headers). In particular, clarify that `io_uring` removes the error callback because it is a pending kernel request, while `epoll` keeps it registered. I'm leaving this comment here to avoid repeating the same concern across both backend declarations.

update: I see some of the documentation is in the cc for epoll / uring. I would move small partof it to  the header. pseudo example:

epoll:
```cpp
// Cancels pending reads and writes and unregisters RegisterOnRecv() callbacks.
// Registered error callbacks remain active.
error_code CancelInFlightOps() final;
```

and uring should say its own extra behavior:

```cpp
// Cancels pending reads and writes and unregisters RegisterOnRecv() and
// RegisterOnErrorCb() callbacks.
error_code CancelInFlightOps() final;
```

util/fibers/epoll_socket.cc
```diff
@@ -193,6 +193,57 @@ auto EpollSocket::Close() -> error_code {
   return ec;
 }
 
+std::error_code EpollSocket::CancelInFlightOps() {
```
glevkovich, line 196: add here:
`DCHECK(proactor() && proactor()->InMyThread());`

util/fibers/uring_socket.cc
```diff
@@ -103,6 +103,44 @@ auto UringSocket::Close() -> error_code {
   return ec;
 }
 
+auto UringSocket::CancelInFlightOps() -> error_code {
```
glevkovich, line 106: Add here
`DCHECK(proactor() && proactor()->InMyThread()); ?`

util/fibers/uring_socket.cc
```diff
@@ -103,6 +103,44 @@ auto UringSocket::Close() -> error_code {
...
+  if (fd_ < 0)
+    return {};
+
+  // The fd-scoped cancel below matches ANY pending request on this fd, not just the write/read
+  // we actually want to unstick - including the error-callback poll from RegisterOnErrorCb()
+  // and the multishot recv from RegisterOnRecv(). Those have their own dedicated teardown paths
+  // that correctly reset this socket's bookkeeping.
+  // Run them first so the broad cancel below doesn't silently swallow one of their completions 
+  // and leave that state stale (e.g. a dangling) error_cb_wrapper_.
```
glevkovich, line 115: nit: you have here whitspace clang-format (trailing space) issue and in few more lines. run clang-format on your PR again.

util/fiber_socket_base.h
```diff
@@ -199,6 +199,14 @@ class LinuxSocketBase : public FiberSocketBase {
...
+  // forcing them to fail immediately instead of waiting to resolve naturally. Unlike
+  // Shutdown()/Close(), this can interrupt an operation that is already in flight - e.g. a
+  // write stalled because the peer isn't draining its receive buffer, which Shutdown()/Close()
+  // cannot force to complete, since io_uring holds its own reference to a submitted operation
+  // until it resolves on its own.
+  virtual error_code CancelInFlightOps() = 0;
```
glevkovich, line 208: I think that since Close (and other API function)  is marked ABSL_MUST_USE_RESULT , This one should mark as well.
Not marking it with ABSL_MUST_USE_RESULT (even if you do not agree with it) is out of scope for this PR.

PR #629, opened 2026-08-06
fix(aws): drop "Bearer " prefix from container-credentials auth header

### LINKED ISSUE #627: Container credentials provider sends malformed Authorization header (Bearer prefix), breaking EKS Pod Identity
### Summary

`AwsCredsProvider::TryContainerCreds()` in `util/cloud/aws/aws_creds_provider.cc` sends the container-credentials auth token with an incorrect `Bearer ` prefix:

```cpp
if (!auth_token.empty()) {
  headers.push_back({"Authorization", absl::StrCat("Bearer ", auth_token)});
}
```

This breaks credential retrieval from the **EKS Pod Identity Agent** (`http://169.254.170.23/v1/credentials`, reached via `AWS_CONTAINER_CREDENTIALS_FULL_URI` + `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`, as injected by the EKS Pod Identity mutating webhook), which expects the **raw token** as the header value, with no `Bearer ` scheme prefix. The agent rejects the malformed request with `HTTP 400`.

### Evidence this is wrong

The official AWS SDKs implement this same "container credentials" protocol (it's shared between ECS task metadata and EKS Pod Identity) and send the token unmodified. E.g. `aws-sdk-go-v2`, `credentials/endpointcreds/internal/client/middleware.go`:

```go
if len(params.AuthorizationToken) > 0 {
    const authHeader = "Authorization"
    request.Header[authHeader] = append(request.Header[authHeader][:0], params.AuthorizationToken)
}
```

No `Bearer` prefix is added anywhere in that path.

### Impact

Any consumer of helio's native AWS S3 client that relies on **EKS Pod Identity** (as opposed to IRSA/web-identity, which uses a different code path — `TryWebIdentity()` — and is unaffected) fails to obtain credentials at all. Downstream in Dragonfly this surfaces as:

```
aws: FetchUrl 169.254.170.23/v1/credentials returned 400
aws: EC2 metadata disabled
aws: no credential provider found
```
and the process fails to initialize S3 snapshot storage (`Failed to initialize AWS S3 snapshot storage: Permission denied`), so the pod never starts when S3 snapshotting is configured.

This is a regression path distinct from `dragonflydb/dragonfly#7666` (that issue's fixes only covered `TryWebIdentity()`/STS) and from `dragonflydb/dragonfly#5172` (EKS Pod Identity support that was added to the now-optional aws-sdk-cpp backend, which is compiled out by default as of `dragonflydb/dragonfly#7945`, so helio's native client is now the only path exercised in default builds).

### Reproduction

1. Deploy a helio-based AWS S3 client (e.g. Dragonfly `v1.40.0`) as a pod on EKS with an **EKS Pod Identity association** (not IRSA) for its ServiceAccount.
2. Observe the EKS Pod Identity webhook injects:
   ```
   AWS_CONTAINER_CREDENTIALS_FULL_URI=http://169.254.170.23/v1/credentials
   AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE=/var/run/secrets/pods.eks.amazonaws.com/serviceaccount/eks-pod-identity-token
   ```
3. On startup, `TryContainerCreds()` reads the token file, sends `Authorization: Bearer <token>` to `169.254.170.23/v1/credentials`.
4. The Pod Identity Agent returns `HTTP 400` (malformed token), credential retrieval fails, and (with IMDS disabled, as is standard EKS practice) the whole provider chain is exhausted.

### Suggested fix

Drop the `"Bearer "` prefix in `TryContainerCreds()`:

```cpp
if (!auth_token.empty()) {
  headers.push_back({"Authorization", auth_token});
}
```

### Environment

- helio: current `master` as vendored into Dragonfly `v1.40.0` (`ghcr.io/dragonflydb/dragonfly:v1.40.0`)
- Platform: EKS, `eks-pod-identity-agent` addon, `terraform-aws-modules/eks-pod-identity/aws` association

kanaevad:
## Summary
- `TryContainerCreds()` sent `Authorization: Bearer <token>` to the ECS/EKS Pod Identity container-credentials endpoint, but that protocol expects the raw token with no scheme prefix. The EKS Pod Identity Agent rejects the malformed header with HTTP 400, so credential retrieval fails whenever a workload relies on Pod Identity instead of IRSA.
- Extracted the header-building logic into `ContainerCredsHeaders()` so the fix has direct unit coverage without needing a real network round-trip.
- IRSA is unaffected — `TryWebIdentity()` is a separate code path that sends its token as a `WebIdentityToken` POST body param, never as an `Authorization` header.

Fixes #627

## Test plan
- [x] Added `ContainerCredsHeadersNoBearerPrefix` / `ContainerCredsHeadersEmptyTokenOmitsHeader` unit tests; verified they fail against the pre-fix code and pass with the fix.
- [x] Built a full `linux/amd64` Dragonfly release image with this fix applied on top of the exact helio commit that `v1.40.0` vendors, and deployed it to a real EKS cluster with a Pod Identity association (no IRSA). Confirmed in logs:
  ```
  aws_creds_provider.cc:484] aws: loaded credentials; provider=container-credentials
  ```
  followed by a successful S3 snapshot load and full sync/takeover.

### DIFF
--- a/util/cloud/aws/aws_creds_provider.cc
+++ b/util/cloud/aws/aws_creds_provider.cc
@@ -252,6 +252,17 @@ string StsEndpoint(string_view region) {
   return "sts.amazonaws.com";
 }
 
+vector<pair<string, string>> ContainerCredsHeaders(string_view auth_token) {
+  vector<pair<string, string>> headers;
+  if (!auth_token.empty()) {
+    // ECS/EKS Pod Identity container-credentials endpoints expect the raw token as the
+    // header value, with no "Bearer " scheme prefix (unlike normal OAuth bearer auth).
+    // See https://github.com/romange/helio/issues/627.
+    headers.push_back({"Authorization", string(auth_token)});
+  }
+  return headers;
+}
+
 AwsCredsProvider::AwsCredsProvider(string region, string endpoint_override)
     : region_(std::move(region)), endpoint_override_(std::move(endpoint_override)) {
 }
@@ -454,11 +465,7 @@ error_code AwsCredsProvider::TryContainerCreds() {
     }
   }
 
-  // Determine if http or https.
-  vector<pair<string, string>> headers;
-  if (!auth_token.empty()) {
-    headers.push_back({"Authorization", absl::StrCat("Bearer ", auth_token)});
-  }
+  vector<pair<string, string>> headers = ContainerCredsHeaders(auth_token);
 
   auto [host, port, path, is_https] = ParseHttpUrl(uri);
   auto resp = FetchUrl(host, port, path, h2::verb::get, headers, "", connect_ms_,
--- a/util/cloud/aws/aws_creds_provider.h
+++ b/util/cloud/aws/aws_creds_provider.h
@@ -4,6 +4,8 @@
 #pragma once
 
 #include <atomic>
+#include <utility>
+#include <vector>
 
 #include "base/RWSpinLock.h"
 #include "util/cloud/utils.h"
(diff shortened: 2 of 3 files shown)


romange:
@kanaevad  please sign your commit with a gpg key submitted to github (verified)

kanaevad:
> @kanaevad please sign your commit with a gpg key submitted to github (verified)

Done

romange (approved):

PR #633, opened 2026-08-09
refactor(tls): complete async I/O boundary

glevkovich:
This PR continues the staged async extraction by narrowing the interface used by the async request state machine.

`TlsAsyncReq` no longer reaches directly into `TlsSocket` or retains a `TlsSocket*`. `TlsAsyncIo` now owns the small operations needed to drive async TLS reads and writes: engine buffer access, upstream I/O dispatch, flag coordination, output accounting, and resuming a blocked request. This keeps async request ownership and progression inside `TlsAsyncIo` while preserving the existing state machine flow and callback behavior.

The test hooks were updated to construct async requests through the same `TlsAsyncIo` boundary.

The PR also stabilizes the affected TLS socket integration tests:
- Wait for the server-side TLS handshake before ordinary post-handshake I/O in `ShortWrite`, `TryRecvVector`, and `SendScatterGather`.
- Treat `EBUSY` as retryable in `TryRecvVector`, matching the recently added `TryRecv` contention result alongside `EAGAIN`/`EWOULDBLOCK`.

No production behavior change is intended.

### DIFF
--- a/util/tls/tls_async_req.h
+++ b/util/tls/tls_async_req.h
@@ -15,17 +15,14 @@ namespace tls {
 
 class TestDelegator;
 class TlsAsyncIo;
-class TlsSocket;
 
 // Drives one TLS read or write request until it invokes its completion callback.
 class TlsAsyncReq {
  public:
   enum Role : std::uint8_t { READER, WRITER };
   // Captures one caller request and the components that advance it.
-  TlsAsyncReq(TlsSocket* owner, TlsAsyncIo* async_io, io::AsyncProgressCb cb, const iovec* v,
-              uint32_t len, Role role)
-      : owner_(owner), async_io_(async_io), caller_completion_cb_(std::move(cb)), vec_(v),
-        len_(len), role_(role) {
+  TlsAsyncReq(TlsAsyncIo* async_io, io::AsyncProgressCb cb, const iovec* v, uint32_t len, Role role)
+      : async_io_(async_io), caller_completion_cb_(std::move(cb)), vec_(v), len_(len), role_(role) {
   }
   TlsAsyncReq(const TlsAsyncReq&) = delete;
   TlsAsyncReq& operator=(const TlsAsyncReq&) = delete;
@@ -42,8 +39,8 @@ class TlsAsyncReq {
 
  private:
   friend class TestDelegator;
+  friend class TlsAsyncIo;
 
-  TlsSocket* owner_;
   TlsAsyncIo* async_io_;
   io::AsyncProgressCb caller_completion_cb_;
   const iovec* vec_;
@@ -67,8 +64,6 @@ class TlsAsyncReq {
   void AsyncReadProgressCb(io::Result<size_t> result);
   // Advances the read or write state machine after an engine operation.
   void AsyncRoleBasedAction();
-  // Resumes the request deferred behind a completed upstream operation.
-  void RunPending();
 };
 
 }  // namespace tls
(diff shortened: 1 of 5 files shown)


util/tls/tls_socket_async.cc
```diff
@@ -314,25 +362,25 @@ void TlsAsyncIo::AsyncWriteSome(const iovec* v, uint32_t len, io::AsyncProgressC
...
     return;
   }
 
-  if (blocked->role_ == Role::WRITER) {
-    auto current = std::move(async_io_->async_write_req_);
-    owner_->AsyncWriteSome(current->vec_, current->len_, std::move(current->caller_completion_cb_));
+  if (blocked->role_ == TlsAsyncReq::WRITER) {
+    auto current = std::move(async_write_req_);
+    AsyncWriteSome(current->vec_, current->len_, std::move(current->caller_completion_cb_));
     return;
```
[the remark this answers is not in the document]
glevkovich: drop. this cannot happen.

util/tls/tls_socket_test.cc
```diff
@@ -554,12 +557,13 @@ TEST_P(TlsSocketTest, TryRecvVector) {
       } else if (res && (*res == 0)) {
         break;  // EOF
       } else if (res.error() != std::errc::resource_unavailable_try_again &&
-                 res.error() != std::errc::operation_would_block) {
-        // Real error occurred (not EAGAIN/EWOULDBLOCK)
+                 res.error() != std::errc::operation_would_block &&
+                 res.error() != std::errc::device_or_resource_busy) {
+        // Fatal error occurred.
         break;
```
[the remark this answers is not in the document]
glevkovich: fixed

util/tls/tls_async_req.h
```diff
@@ -42,8 +39,8 @@ class TlsAsyncReq {
 
  private:
   friend class TestDelegator;
+  friend class TlsAsyncIo;
```
kostasrim, line 42: what's wrong with public api's ?

util/tls/tls_socket_async.cc
```diff
@@ -22,21 +22,102 @@ TlsAsyncIo::TlsAsyncIo(TlsSocket* owner) : owner_(owner) {
...
+  DVLOG(2) << "Engine::Read tried to read " << v->iov_len << " bytes, got " << op_val;
+  return op_val;
+}
+
+TlsAsyncIo::PushResult TlsAsyncIo::PushUserDataToEngine(const iovec* v, uint32_t len) {
+  TlsSocket::PushResult result = owner_->PushUserDataToEngine(v, len);
+  return {result.written, result.engine_opcode};
+}
+
+size_t TlsAsyncIo::EngineOutputPending() const {
```
kostasrim, line 36: Not a big fan of those single line proxies and IMO it's hard for me to understand the benefits of this separation (general comment about the refactoring) but whatever works 🤷

kostasrim (approved):

util/tls/tls_async_req.h
glevkovich: This is a good argument. Nothing is fundamentally wrong with making this the public API, but I kept it private with friendship because these methods are coordination details rather than API intended for general callers, but I agree that does not buy much here.

There is one useful part of the extraction: TlsAsyncIo owns the active request state. The forwarding methods are mainly there to stop TlsAsyncReq from reaching into TlsSocket directly, but I agree they currently make the separation feel mechanical or "artificial" rather than meaningful.

util/tls/tls_socket_async.cc
glevkovich: I see your point. The main goal here was to move the async request ownership out of TlsSocket. These small forwarding methods stop TlsAsyncReq from directly touching TlsSocket, but they do not really remove the coupling yet.

I kept it this way to make the refactor small and avoid changing behavior. I agree the one-line methods are not ideal and feel "artificial". I prefer to leave this approved PR as-is and improve that separation later, instead of adding more changes here.

PR #638, opened 2026-08-10
feat: run pre-commit as part of the ci

kostasrim:
We did not run `pre-commit` and we never checked if the code is actually formatted 🤦

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,27 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-commit:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: actions/cache@v4
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit-${{ runner.os }}-${{ hashFiles('.pre-commit-config.yaml') }}
+      - uses: pre-commit/action@v3.0.1
+        with:
+          extra_args: >-
+            --show-diff-on-failure --color=always
+            --from-ref ${{ github.event.pull_request.base.sha }}
+            --to-ref ${{ github.event.pull_request.head.sha }}
+
   build:
     name: ${{ matrix.config.name }} on (${{ matrix.os.arch }})
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -14,6 +14,27 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-commit:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
```
glevkovich: actions/checkout@v4 is old, use the latest:
actions/checkout@v7

.github/workflows/ci.yml
```diff
@@ -14,6 +14,27 @@ concurrency:
...
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v5
```
glevkovich: use actions/setup-python@v7

.github/workflows/ci.yml
```diff
@@ -14,6 +14,27 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-commit:
+    if: github.event_name == 'pull_request'
```
glevkovich, line 18: This can have security issues.

Set this to restrict the job to be able to read code only:
```
if: github.event_name == 'pull_request'
    permissions:
      contents: read
```

.github/workflows/ci.yml
```diff
@@ -14,6 +14,27 @@ concurrency:
...
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
glevkovich, line 25: You dont want to keep the git token for action checkout, or very nasty things can happen if a malicious user reused it.
set persist-credentials:

```
- uses: actions/checkout@v7
        with:
          fetch-depth: 0
          persist-credentials: false
```

.github/workflows/ci.yml
```diff
@@ -14,6 +14,27 @@ concurrency:
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - uses: actions/cache@v4
```
glevkovich: you dont need all this actions/cache block, you should remove it.

pre-commit/action@v3.0.1 already restores and saves ~/.cache/pre-commit, using a key that also includes env.pythonLocation.So it already has a built in cache. So you'd better remove it to avoid cache actions managing the same directory with different keys

kostasrim:
@glevkovich PTAL

glevkovich (dismissed):
LGTM,
 you forgot one comment to address:
https://github.com/romange/helio/pull/638/changes#r3755410320

upgrade actions/setup-python@v5 to @v7.

glevkovich (approved):
LGTM

PR #640, opened 2026-08-11
Make function param name match to help clangd and IDEs

abhijat:
Purely cosmetic change, helps certain IDEs depending partially on clangd (like CLion Nova). By matching the parameter name in declaration vs definition suppresses false positives on `AwaitBrief` throughout dragonfly


glevkovich (approved):
LGTM

kostasrim (approved):

PR #641, opened 2026-08-19
chore: rename irq-based naming with tail

kostasrim:
Rename to address the comment in https://github.com/romange/helio/pull/628#discussion_r3778478739

`proactor_irq_cpus` to `proactor_tail_cpus`

### DIFF
--- a/util/fibers/proactor_pool.cc
+++ b/util/fibers/proactor_pool.cc
@@ -28,11 +28,12 @@ using namespace std;
 
 ABSL_FLAG(uint32_t, proactor_threads, 0, "Number of io threads in the pool");
 ABSL_FLAG(string, proactor_affinity_mode, "on", "can be on, off or auto");
-ABSL_FLAG(string, proactor_irq_cpus, "",
-          "Explicit list of cpu ids/ranges handling irqs, e.g. \"1,4,6,7\" or \"0-47\". These "
-          "cpus are pinned to the highest proactor thread indices; every other online cpu is "
-          "pinned to the lower indices. Does not affect pool size. Ignored (with a logged error) "
-          "if it names an offline cpu or more cpus than the pool has threads.");
+ABSL_FLAG(string, proactor_tail_cpus, "",
+          "Explicit list of cpu ids/ranges to pin at the end of the thread list, e.g. "
+          "\"1,4,6,7\" or \"0-47\". These cpus are pinned to the highest proactor thread "
+          "indices; every other online cpu is pinned to the lower indices. Does not affect "
+          "pool size. Ignored (with a logged error) if it names an offline cpu or more cpus "
+          "than the pool has threads.");
 
 namespace util {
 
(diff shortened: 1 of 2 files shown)


kostasrim:
@romange let me know if you want to follow a different naming convention :)

util/fibers/proactor_pool.cc
```diff
@@ -28,11 +28,12 @@ using namespace std;
 
 ABSL_FLAG(uint32_t, proactor_threads, 0, "Number of io threads in the pool");
 ABSL_FLAG(string, proactor_affinity_mode, "on", "can be on, off or auto");
-ABSL_FLAG(string, proactor_irq_cpus, "",
-          "Explicit list of cpu ids/ranges handling irqs, e.g. \"1,4,6,7\" or \"0-47\". These "
-          "cpus are pinned to the highest proactor thread indices; every other online cpu is "
-          "pinned to the lower indices. Does not affect pool size. Ignored (with a logged error) "
-          "if it names an offline cpu or more cpus than the pool has threads.");
+ABSL_FLAG(string, proactor_tail_cpus, "",
+          "Explicit list of cpu ids/ranges to pin at the end of the thread list, e.g. "
```
romange, line 32: Ok, now I understand semantics. Are you sure it's enough for machines with several numa nodes? Did you check it?

romange (approved):

util/fibers/proactor_pool.cc
kostasrim: Linux allows to pin irqs to cpu's on each numa node (and ofc that device must be local). So yes it should work as long the instance is configured properly. That being said, no I did not spawn a 3 node beast but saying this I will follow up this evening and post here!

p.s. do not merge this until then

kostasrim: @romange I double checked on c5.18xlarge which has 2 numa nodes and no issue there. I wanted to test on a 4 numa node instance (hpc6a.48xlarge family) but neither dev nor prod has them (so maybe we limit them) 🤷

romange: you can try on GCP with c3-highcpu-176 or c2-standard-60

kostasrim: Happy to!

kostasrim: @romange I used c3-highcpu-176 which has 4 numa nodes. I overwrote the affinity such that each irq is pinned to a single CPU within a numa node and then asked claude to rewrite the test and verify that the CPU's passed in the flag end up the last proactor indices -- no issue. 

I want to point out something just to be clear here. There is a `single nic` card on this node. The way I pinned a CPU from a numa node to handle an irq is non optimal and we should not do this in production. Why ? Because the interrupts arrive in node1 but the packet arrives in the buffers allocated in node0 (where the nic lives). This is a remote memory access and it's penalized/non optimal. I will make sure our prod scripts configure this properly.

kostasrim: I will follow up with a script tomorrow for our prod image and then rerun the test 😄

romange: No need to rerun the tedt. 

I just wanted to know if nic is bound to a single node. Do you know cpu range on that node that was responsible for nic?

kostasrim: I would lie if I told you I remember its affinity 😄 I will follow up with a script this evening though so I will do a quick experiment again.

I am actually curious if aws/gcp offers multi numa node sockets with multiple nic's attached 🤷  (happy to share any information around it)

PR #645, opened 2026-08-26
chore: timeout support to http client

kostasrim:
* add support for http client timeout (jwt authentication will need it)


romange:
@kostasrim  what is jwt authentication?

kostasrim:
> @kostasrim  what is jwt authentication?

Cp wants to add token based authentication to dragonfly:

So user can do: auth user token

https://www.jwt.io for more info

We can sync internally for the rest and how it works

romange (approved):

